PR #4, opened 2026-05-06
Harden file store writes against symlink swaps

sjf:
## Summary

This hardens `fileStore`, `fileStoreSync`, and the advanced `copyIntoRoot()` helper so store writes no longer commit through a caller-derived absolute destination path after validation.

The public API stays the same. Some write failures now surface lower-level `FsSafeError` codes such as `path-alias`, `path-mismatch`, or `not-file` depending on which guard rejects the path.

Related audit report gist: https://gist.github.com/sjf/dbca7178ef63782609f094acec33e164

## Vulnerability

The affected write paths resolved and validated a requested relative path under the store root, but later created parents and committed writes with absolute path strings derived from that earlier result.

That made symlink-parent escapes possible:

1. A caller writes to a path such as `link/write.txt` under the store root.
2. `root/link` is, or becomes, a symlink to a directory outside the store root.
3. The old implementation follows `root/link/...` during the write/temp/rename phase and creates the file outside the store.

For the async store, a same-call race could exploit the gap by swapping `root/link` from an in-tree directory to a symlink after validation but before the final write/copy commit. After pulling latest `main`, the newly added `fileStoreSync()` path had the same static symlink-parent escape for `write`, `writeText`, and `writeJson`.

## Minimal Exploit

```js
import fs from "node:fs/promises";
import path from "node:path";
import os from "node:os";
import { Readable } from "node:stream";
import { copyIntoRoot, fileStore, fileStoreSync } from "@openclaw/fs-safe/store";

const base = await fs.mkdtemp(path.join(os.tmpdir(), "fs-safe-poc-"));
const storeRoot = path.join(base, "root");
const outside = path.join(base, "outside");
const source = path.join(base, "source.txt");

await fs.mkdir(storeRoot, { recursive: true });
await fs.mkdir(outside, { recursive: true });
await fs.writeFile(source, "pwned");
await fs.symlink(outside, path.join(storeRoot, "link"), "dir");

const store = fileStore({ rootDir: storeRoot });
const syncStore = fileStoreSync({ rootDir: storeRoot });

await store.write("link/write.txt", "pwned");
await store.writeStream("link/stream.txt", Readable.from(["pwned"]));
await store.copyIn("link/copy.txt", source);
await copyIntoRoot({
  rootDir: storeRoot,
  relativePath: "link/advanced-copy.txt",
  sourcePath: source,
});

syncStore.write("link/sync-write.txt", "pwned");
syncStore.writeText("link/sync-text.txt", "pwned");
syncStore.writeJson("link/sync-json.json", { pwned: true });

console.log(await fs.readdir(outside));
// Vulnerable behavior: files are created in `outside`.
// Fixed behavior: operations reject and `outside` remains [].
```

## Fix

The async store write paths now commit through the root-pinned machinery:

- `fileStore.write()` opens a `Root`, creates the parent via `Root.mkdir()`, and commits with `Root.write()`.
- `fileStore.copyIn()` and `copyIntoRoot()` open a `Root`, create the parent via `Root.mkdir()`, and commit with `Root.copyIn()`.
- `fileStore.writeStream()` streams directly through the pinned write helper using the already-opened root identity and parent-relative target, then verifies the resulting file identity with `Root.open()`.
- `writeStream()` still accepts the same `Readable` API and does not stage the stream into a separate temp source file.

The sync store now rejects symlinked/non-directory parent components before writing:

- `fileStoreSync.write()`, `writeText()`, and `writeJson()` walk/create parent directories with `lstatSync()` and `realpathSync()` under the store root.
- The sync path rejects any symlink parent component and verifies that the parent realpath has not changed before the final rename.

## Regression Coverage

Added POSIX regressions for:

- pre-existing symlink parent escapes across `fileStore.write`, `writeStream`, `copyIn`, `copyIntoRoot`, `fileStoreSync.write`, `writeText`, and `writeJson`
- same-call symlink parent swaps during async `write`, `writeStream`, `copyIn`, and `copyIntoRoot`

## Verification

- `corepack pnpm run build`
- `corepack pnpm exec vitest run test/new-primitives.test.ts test/api-coverage.test.ts test/additional-bypass-parity.test.ts`
- `corepack pnpm test` outside the sandbox: 22 test files, 221 tests passed
- `git diff --check`
- built-package PoC replay: async `write`, `writeStream`, `copyIn`, and `copyIntoRoot` rejected with `path-alias`; sync `write`, `writeText`, and `writeJson` rejected with `not-file`; the outside directory stayed `[]`


steipete:
Thanks Sarah. I verified the report against current main and landed a maintainer replacement fix in a6149c2.

What landed:
- non-private fileStore.write(), writeStream(), and copyIn() now commit through the root-pinned write/copy path instead of caller-derived absolute sibling-temp destinations
- fileStoreSync.write(), writeText(), and writeJson() now reject symlinked parent components and re-check the parent directory during the sync write path
- added regression coverage for async + sync store writes through symlink parents
- replayed the PoC against the rebuilt package: all six operations rejected and the outside directory stayed empty

Proof on main:
- pnpm check: 23 files / 228 tests passed
- pnpm docs:site passed
- GitHub ci and coverage are green for a6149c2

Closing this PR as superseded by the landed main commit. Appreciate the security report and the detailed repro.

PR #5, opened 2026-05-06
test: expand boundary bypass coverage

jesse-merhi:
## Summary

- rename bypass tests away from OpenClaw/parity framing to boundary-focused names
- add additional helper boundary tests for archive, temp, local file URL, install path, walk, trash, and JSON store surfaces
- add adversarial generated-payload tests across read/write/open/remove/move/copy surfaces
- exercise dot-segment matrices, encoded/double-encoded separators, Windows/backslash/drive/UNC-looking payloads, control characters, symlink chains, and outside-clobber assertions
- make the existing pinned-copy source-swap test deterministic on Linux by swapping with a pre-created replacement file

## Validation

- `pnpm check`

## Notes

This PR is intentionally test-focused. The generated adversarial corpus allows either safe rejection or literal in-root handling, but always asserts outside files remain untouched.

### DIFF
--- a/test/additional-boundary-bypass.test.ts
+++ b/test/additional-boundary-bypass.test.ts
@@ -1,0 +1,189 @@
+import fs from "node:fs";
+import fsp from "node:fs/promises";
+import os from "node:os";
+import path from "node:path";
+import { pathToFileURL } from "node:url";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { safeFileURLToPath } from "../src/local-file-access.js";
+import { assertCanonicalPathWithinBase, resolveSafeInstallDir } from "../src/install-path.js";
+import { createJsonStore } from "../src/json-document-store.js";
+import { movePathToTrash } from "../src/trash.js";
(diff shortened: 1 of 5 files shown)


steipete:
Thanks Jesse. I looked through this after fixing #4 and salvaged the useful part into main in a6149c2.

What landed from this PR:
- a corrected generated adversarial boundary corpus in test/adversarial-boundary-payloads.test.ts
- the new corpus is included in pnpm test:security
- coverage now exercises traversal/control/Windows-looking payloads across open/read/stat/write/create/append plus copy/move clobber checks

I did not take the PR wholesale because it now conflicts with current main and some helper-level assertions were less exact than the current test helpers: the archive staging call used the wrong parameter shape, and the walk follow check was lexical rather than realpath-based. The landed version keeps the strong corpus signal without those weaker assertions.

Proof on main:
- pnpm check: 23 files / 228 tests passed
- pnpm test:security: 55 security tests passed
- GitHub ci and coverage are green for a6149c2

Closing as superseded by the landed main commit. Appreciate the test-focused pass.

PR #7, opened 2026-05-06
[codex] add safe external output writer

jesse-merhi:
## Summary
- Add `writeExternalFileWithinRoot()` for external libraries that require a filesystem output path.
- Stage writes in private temp storage, then finalize through `Root.copyIn()` under the requested root.
- Document the new `@openclaw/fs-safe/output` subpath and add regression coverage for traversal, symlink parent escapes, hardlink aliases, cleanup, and size/mode plumbing.

## Validation
- `pnpm test test/output.test.ts`
- `pnpm build`
- `pnpm check`
- `pnpm docs:site`
- `git diff --check HEAD~1..HEAD`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixes
 
+- Add `writeExternalFileWithinRoot()` for libraries that require an output path while preserving caller-provided destination names. (#7; thanks @jesse-merhi)
 - Reject `fileStore()` and `fileStoreSync()` writes through symlinked parent directories so store commits cannot escape the configured root.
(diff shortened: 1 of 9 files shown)


steipete:
Landed via rebase onto main.

- Gate: `pnpm test test/output.test.ts`; `pnpm check`; `pnpm docs:site`; `git diff --check`
- PR fix commit: d144817e7cdb11bcc8c582da28b6dcda46244e80
- Landed main commit: b57002a6a1c7ecbb2f21344da19e58d85294612b

Thanks @jesse-merhi!

PR #19, opened 2026-05-18
fix(json): retry async JSON reads when atomic rename races

yetval:
## Summary

- Tag the regular-file identity-change throw as `FsSafeError("path-mismatch")` instead of a raw `Error`, matching the sentinel already used in `src/file-store.ts:149`, `src/directory-guard.ts:35`, `src/pinned-write.ts:292`, and `src/pinned-python.ts:514` for the same temp-file + rename race semantic.
- Wrap `readJson`, `readJsonIfExists`, and `tryReadJson` in a bounded retry (max 3 attempts, 50ms exponential backoff) that only retries on `path-mismatch`. Parse errors and other read failures still fail fast on the first attempt, so corruption is not masked.
- Add a real-disk regression in `test/json.test.ts` that spawns a concurrent `writeTextAtomic` loop against a `readJson` loop on a `mkdtemp` scratch file and asserts zero `raceErrors` over a 1-second window. Keep one mocked test for the retry-budget exhaustion branch (hard to hit deterministically on real disk; `docs/contributing.md` allows `vi.mock` sparingly for cases like this).

## Why

`openclaw/openclaw` (and any other consumer of `@openclaw/fs-safe/json`) is hitting `JsonFileReadError: Failed to read JSON file: .../paired.json` under normal multi-client operation. The writer uses temp-file + `rename` (atomic, by design). Concurrent readers race: `statRegularFile` reads the pre-rename inode, `fs.open` then resolves to the post-rename inode, `verifyStableReadTarget` detects the identity mismatch in `src/regular-file.ts:174-179` and throws.

That throw is correct - the read genuinely landed on a different file - but it is a transient state, not corruption. Today the only error type bubbled up is a generic `Error`, so callers cannot retry without doing fragile message matching. The fix typifies it with the existing `path-mismatch` code and adds the bounded retry the issue asks for, scoped to the async public readers.

Tracks openclaw/openclaw#83657.

## Scope notes

- Sync readers (`readJsonSync`, `readRootJsonSync`, `tryReadJsonSync`) are intentionally unchanged - they cannot `await` and the issue's pseudocode is async. A sync retry on top of `Atomics.wait` or a tight spin loop is a separate decision and can land in a follow-up if anyone hits the race on the sync paths.
- `readRegularFile`/`readRegularFileSync` themselves are unchanged in behavior; only the error type the race emits is tightened. Callers that previously string-matched on the message still match the same message (`File changed during read: <path>`).

## Real behavior proof

Real disk, no `fs.open` mock. The new vitest case `recovers readJson from a concurrent real atomic rewrite` runs concurrent `readJson` and `writeTextAtomic` loops against a `mkdtemp` scratch file for 1 second and counts `raceErrors` by matching the cause message.

Against `upstream/main` (`src/json.ts` + `src/regular-file.ts` reverted to upstream, new tests in place):

```
$ pnpm test test/json.test.ts
 × json file helpers > recovers readJson from a concurrent real atomic rewrite 1014ms
 × json file helpers > surfaces JsonFileReadError when read races exceed retry budget 15ms

FAIL  json file helpers > recovers readJson from a concurrent real atomic rewrite
AssertionError: expected 79 to be +0
 ❯ test/json.test.ts:314
    312|     expect(writes).toBeGreaterThan(10);
    313|     expect(okReads).toBeGreaterThan(10);
    314|     expect(raceErrors).toBe(0);

FAIL  json file helpers > surfaces JsonFileReadError when read races exceed retry budget
AssertionError: expected 1 to be greater than or equal to 3

 Test Files  1 failed (1)
      Tests  2 failed | 12 passed (14)
```

79 real `JsonFileReadError`s caused by `File changed during read` during a 1-second concurrent run - exact symptom from openclaw/openclaw#83657. The second failure also pins down today's behavior as "one attempt, no retry."

With the fix:

```
$ pnpm test test/json.test.ts
 Test Files  1 passed (1)
      Tests  14 passed (14)

$ pnpm test
 Test Files  35 passed (35)
      Tests  383 passed | 2 skipped (385)

$ pnpm build
$ pnpm lint:fs-boundary
$ pnpm lint:file-size
```

All clean. The same real-disk concurrent loop that produced 79 race errors against upstream produces zero against the patched build, without `parseErrors` or `otherErrors`, confirming the retry is not masking corruption or unrelated failures.

What was not tested: a real multi-process Discord + Control UI + CLI reproduction against `openclaw/openclaw`'s gateway. The race lives in the kernel rename path, so an in-process concurrent test exercises the same code; a multi-process reproduction behaves the same.

## Verification

- `pnpm test test/json.test.ts`
- `pnpm test`
- `pnpm build`
- `pnpm lint:fs-boundary`
- `pnpm lint:file-size`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,16 @@
 
 ## Unreleased
 
+### Security and Correctness
+
+- Retry async JSON reads (`readJson`, `readJsonIfExists`, `tryReadJson`) up to three times with 50ms exponential backoff when the file is rotated mid-read by an atomic rename, and tag the underlying race as `FsSafeError("path-mismatch")` so callers can distinguish transient swaps from corruption.
+
 ## 0.2.7 - 2026-05-20
 
 ### Security and Correctness
 
 - Restore best-effort Node write fallbacks when the Python helper is disabled or unavailable, preserving the `mode: "off"` contract while documenting the weaker POSIX same-UID race resistance compared with fd-relative helper commits.
 - Harden DeepSec-reported archive staging and input pinning, secret and queue hardlink rejection, absolute output file validation, sidecar lock read failures, ClawSweeper dispatch trust checks, and scoped path defaults.
-- Retry async JSON reads (`readJson`, `readJsonIfExists`, `tryReadJson`) up to three times with 50ms exponential backoff when the file is rotated mid-read by an atomic rename, and tag the underlying race as `FsSafeError("path-mismatch")` so callers can distinguish transient swaps from corruption.
 
 ## 0.2.6 - 2026-05-17
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 - Restore best-effort Node write fallbacks when the Python helper is disabled or unavailable, preserving the `mode: "off"` contract while documenting the weaker POSIX same-UID race resistance compared with fd-relative helper commits.
 - Harden DeepSec-reported archive staging and input pinning, secret and queue hardlink rejection, absolute output file validation, sidecar lock read failures, ClawSweeper dispatch trust checks, and scoped path defaults.
+- Retry async JSON reads (`readJson`, `readJsonIfExists`, `tryReadJson`) up to three times with 50ms exponential backoff when the file is rotated mid-read by an atomic rename, and tag the underlying race as `FsSafeError("path-mismatch")` so callers can distinguish transient swaps from corruption.
 
 ## 0.2.6 - 2026-05-17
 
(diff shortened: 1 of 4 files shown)


yetval:
@clawsweeper re-review

Added real-fs proof. New `scripts/race-repro.mjs` runs concurrent `readJson` against `writeTextAtomic` on a real temp file — no `fs.open` mocks. Measured numbers in updated PR body:

- upstream/main: 521 raceErrors across 247 atomic writes
- this branch: 0 raceErrors across 264 atomic writes

Parse errors and other errors stay at 0 in both runs, so the retry is not masking corruption.

PR #28, opened 2026-06-10
fix: fall back to copy on EPERM/EEXIST rename errors

liuxingwei0601:
## Problem

`movePathWithCopyFallback` only catches `EXDEV` (cross-device) errors from `guardedRename`. On Windows, files locked by antivirus, other processes, or permission issues throw `EPERM`/`EEXIST`, causing install and update flows to fail unnecessarily.

## Root Cause

The `renameWithRetry` helper already has `copyFallbackOnPermissionError` support that handles EPERM/EEXIST, but `movePathWithCopyFallback` calls `guardedRename` directly, bypassing that logic.

## Fix

Catch `EPERM` and `EEXIST` alongside `EXDEV` in `movePathWithCopyFallback` so the copy fallback path handles Windows permission/lock errors.

## Tests

Added two new test cases:
- `falls back to copy on EPERM (Windows file lock)`
- `falls back to copy on EEXIST (destination locked)`

Both pass with `vitest run`.

## Context

This is the same issue that the local EPERM patch addressed in OpenClaw. The fix belongs in `@openclaw/fs-safe` so all consumers benefit.

### DIFF
--- a/src/move-path.ts
+++ b/src/move-path.ts
@@ -261,7 +261,11 @@ export async function movePathWithCopyFallback(
     await guardedRename({ from: options.from, to: options.to });
     return;
   } catch (error) {
-    if ((error as NodeJS.ErrnoException | null)?.code !== "EXDEV") {
+    const code = (error as NodeJS.ErrnoException | null)?.code;
+    // EXDEV = cross-device move (same filesystem copy fallback)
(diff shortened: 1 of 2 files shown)


liuxingwei0601:
Superseded by upstream fix e9e692a2 which already handles Windows EPERM with proper platform scoping. Closing.

PR #29, opened 2026-06-15
feat(walk): report unreadable directories in failedDirs

amknight:
## Summary

`walkDirectory()` / `walkDirectorySync()` swallow `realpath`/`readdir` errors at **every** level — including the walk root — with a bare `catch { return; }`, then return a partial or empty `entries` list. A caller that reconciles destructive state from the listing (stale-row pruning, exported-corpus reconciliation, etc.) cannot tell a transient `EIO`/`EACCES`/`ESTALE` read failure from a genuinely empty directory, so a single blip looks like a mass deletion.

This adds **`WalkDirectoryResult.failedDirs`**: every directory the walk could not enumerate is recorded with its absolute `path`, root-relative `relativePath`, `depth`, and the thrown `error`. Callers can now classify a benign missing-directory race (`ENOENT`) versus a real read failure and skip destructive work when the scan was not authoritative.

Behaviour is otherwise unchanged — unreadable directories are still skipped, and entries from the readable parts of the tree are still returned. The field is additive; existing consumers that ignore it are unaffected.

```ts
type WalkDirectoryFailure = { path: string; relativePath: string; depth: number; error: unknown };
type WalkDirectoryResult = { entries; scannedEntryCount; truncated; failedDirs: WalkDirectoryFailure[] };
```

## Motivation

OpenClaw's memory index treats a directory listing as the source of truth for destructive maintenance (stale-row pruning of its search index). Because `walkDirectory` swallows `realpath`/`readdir` errors, a transient `EIO`/`EACCES`/`ESTALE` blip currently yields an empty/partial listing that the prune treats as mass deletion — a live index-wipe hazard on `main`.

A pending OpenClaw fix guards this by adding an explicit `readdir` probe on each scan root *before* walking it, but that costs a second full `readdir` per root on networked filesystems and still only covers the root, leaving sub-tree read failures invisible. With `failedDirs`, that fix can detect an incomplete scan straight from the walk result — no probe, one `readdir` per directory — and catch failures anywhere in the tree, not just at the root.

## What changed

- `src/walk.ts`: new `WalkDirectoryFailure` type and `failedDirs` on `WalkDirectoryResult`; both walkers record the `realpath`/`readdir` catch sites via a shared `recordFailedDir` helper. Symlink-target kind-resolution failures are intentionally **not** recorded (they are not directory-enumeration failures).
- `test/new-primitives.test.ts`: unreadable sub-tree (readable siblings preserved + failure reported with `EACCES`), unreadable walk root (empty listing + reported), the synchronous walker, and the empty-`failedDirs` contract for a fully readable tree.
- `docs/walk.md`, `README.md`: document `failedDirs` and the incomplete-vs-empty distinction for pruning callers.

## Verification

Run on macOS, Node 25, `pnpm check` (lint:file-size + lint:fs-boundary + build + test):

- `pnpm check` — **green**: both lints pass, `tsc` builds clean, **412 passed | 3 skipped (415)** across 37 files.
- New directory-walking tests: **6 passed** (4 new + 2 pre-existing), and confirmed RED first (each new test failed with `failedDirs` undefined before the implementation).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,6 @@
 
 ## Unreleased
 
-### Features
-
-- Add `WalkDirectoryResult.failedDirs` to `walkDirectory`/`walkDirectorySync`, recording every directory whose `realpath`/`readdir` threw so destructive callers can tell an incomplete scan from an empty directory instead of treating a transient `readdir` failure as mass deletion. (#29; thanks @amknight)
-
 ### Compatibility
 
 - Require Node.js 22 or newer for the npm package and docs, matching the maintained CI matrix.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+- Add `WalkDirectoryResult.failedDirs` to `walkDirectory`/`walkDirectorySync`, recording every directory whose `realpath`/`readdir` threw so destructive callers can tell an incomplete scan from an empty directory instead of treating a transient `readdir` failure as mass deletion. (#29; thanks @amknight)
+
 ### Compatibility
 
 - Require Node.js 22 or newer for the npm package and docs, matching the maintained CI matrix.
--- a/README.md
+++ b/README.md
@@ -371,7 +371,7 @@ for (const file of scan.entries) {
 }
 ```
 
-Check `scan.truncated` before treating the result as complete.
+Check `scan.truncated` before treating the result as complete, and `scan.failedDirs` to tell an incomplete scan (a directory that could not be read) from an empty one before pruning state from the listing.
 
 ## Archive extraction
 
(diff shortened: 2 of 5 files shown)


steipete:
Maintainer fixup pushed in 178122f:

- preserves source compatibility for callers that manually construct `WalkDirectoryResult`
- keeps `failedDirs` required on values inferred from both walker functions
- removes the release-owned changelog hunk
- adds a regression test for the legacy result shape

Validation:
- `pnpm check`: 37 test files, 414 passed, 3 skipped
- generated declaration inspection
- autoreview: no actionable findings

steipete:
Landed as b38f977a401cced867c6d0270f5c7de44b5382bb. Thanks @amknight.

The walkers now expose unreadable-directory failures without breaking legacy `WalkDirectoryResult` constructors. Full local checks passed, the complete Node 22/24 macOS/Linux/Windows matrix is green after one unrelated flaky-job rerun, and autoreview found no actionable issues.

PR #33, opened 2026-07-01
fix(write): add renameIdentity option for FUSE mounts with unstable inode numbers

### LINKED ISSUE #32: Post-rename identity check (sameFileIdentity) always fails on rclone FUSE mounts, even with zero concurrency
## Summary

`runPinnedWriteFallback`'s post-rename identity check (`sameFileIdentity` comparing `(dev, ino)`) always fails on rclone FUSE mounts, even with a single process and zero concurrency. The filesystem creates a fresh inode number on every path lookup, including a lookup that happens right after we renamed a file we wrote ourselves. Any overwrite-write through `writeFileInRoot`/`runPinnedWriteFallback` against a root on this kind of mount doesn't work.

I hit this through `openclaw wiki status`, which throws `FsSafeError: fallback target changed during write` on every run against a `memory-wiki` vault stored on an rclone-mounted Google Drive workspace.

## How this differs from #92134 / #85876 / #83740

Those are intermittent `path-mismatch` failures from a real concurrent rename racing a read, and they go away on retry. This doesn't match that pattern since it is one process without concurrency that fails every time. There's no race window to retry past, because the filesystem's `rename(2)` doesn't preserve inode identity the way `sameFileIdentity` expects.

## Root cause

`src/pinned-write.ts`, `runPinnedWriteFallback`:

```ts
tempStat = await handle.stat();                 // identity via held fd, before rename
...
await fs.rename(tempPath, targetPath);
targetStat = await fs.lstat(targetPath);         // fresh path lookup, after rename
if (targetStat.isSymbolicLink() || !sameFileIdentity(targetStat, expectedTempStat)) {
  throw new FsSafeError("path-mismatch", "fallback target changed during write");
}
```

`sameFileIdentity` (`src/file-identity.ts`) requires `left.ino === right.ino` plus matching `dev`. That only holds if `rename(2)` preserves the source's inode at the destination which is true on POSIX filesystems but not guaranteed elsewhere.

### Reproduction (rclone mount, `--vfs-cache-mode writes --dir-cache-time 1m`, Google Drive remote)

Plain `mv` inside the mounted directory, run 5 times in a row, nothing else touching the file:

```
run 1: pre-rename ino=16151710212077576736  post-rename ino=3761490782177196776
run 2: pre-rename ino=16151713510612461369  post-rename ino=3761494080712081409
run 3: pre-rename ino=16151712411100833158  post-rename ino=3761492981200453198
run 4: pre-rename ino=16151715709635717791  post-rename ino=3761496279735337831
run 5: pre-rename ino=16151714610124089580  post-rename ino=3761495180223709620
```

As a control, I ran the same `mv` test on the container's `overlay` root filesystem, which isn't FUSE:

```
pre-rename ino=186377  post-rename ino=186377   (matches)
```

I also checked whether this was general cache churn unrelated to rename. Stat'ing an unrenamed, unchanged file twice, 2 seconds apart, on the same rclone mount gives the same inode both times. The instability is specific to rename, not the mount being flaky in general.

## I looked for a way to keep the same TOCTOU protection on this kind of filesystem, and couldn't find one

Before reporting this I wanted to check whether there's some inode-independent way to get the same guarantee on a mount like this. One obvious idea was to keep the file descriptor open across the rename instead of closing it first, which is what the code does today, and compare `fstat()` on that same fd before and after, but that didn't address the check's goal.

```
fd9 identity before rename (held fd):          ino=17015438226736127255
fd9 identity after rename  (same held fd):     ino=17015438226736127255   (matches)
fresh lookup of target by path after rename:   ino=7749167912470427203    (different)
```

A held fd does stay stable across the rename. But that doesn't actually buy you anything. To check whether `targetPath` still points at what you wrote, you have to look it up by name again, and on this mount every fresh lookup gets a new inode no matter what. Comparing the held fd against itself would always pass, even in the exact attacker-swap scenario the check is meant to catch. As far as I can tell, this filesystem doesn't have any way to ask whether a name currently points at the same object as a given fd.

I also checked whether rclone has a flag for stable inodes. `rclone mount --help` has nothing for this. Rclone creates inode numbers from a process-local counter (`vfs.newInode()`), not anything tied to a stable remote file ID, so I doubt rclone could fix this on its end.

## Previous context on the check

This is not an ask to weaken or drop the intent behind `sameFileIdentity`. From `docs/security-model.md` and the `66bf772` commit ("harden deepsec-reported filesystem races"), this check catches a same-UID racer swapping the rename target between `rename()` and the verifying `lstat()`.

Instead I'm suggesting using hashing to compare that the written content matches the expectation.

## Environment

- `openclaw` 2026.6.10 (aa69b12), bundling `@openclaw/fs-safe` 0.3.0
- Filesystem: `rclone mount <remote>: /data --vfs-cache-mode writes --dir-cache-time 1m --poll-interval 15s` (Google Drive remote), `fuse.rclone`
- Linux, Docker container, single process accessing the mount the whole time

## Suggested next step

At minimum this seems worth a line in `docs/security-model.md` / `docs/root.md` saying the overwrite path in `runPinnedWriteFallback` needs rename to preserve inode identity, and that it won't work at all, not just be more race-prone, on FUSE mounts that don't do that. Rclone is a confirmed example. Happy to dig up more detail if it helps.

jlautman:
## Summary

- Add explicit `renameIdentity: "verify-content-with-lock"` compatibility writes for rclone-style FUSE mounts whose destination inode changes across rename.
- Keep strict `(dev, ino)` verification as the unchanged default.
- Verify matching content under a cooperative lock only for the rename boundary; keep subsequent path identity checks strict.

Closes #32.

## Security contract

This is an explicit weaker opt-in for trusted/cooperating writers. It does not prove that the destination still names the temp-file object, retain the Python helper's fd-relative parent pinning, or stop a same-UID writer that ignores the advisory lock.

The compatibility lock is a per-target, SHA-256-keyed sidecar stored in the already-canonical root. That avoids creating a lock through a missing or raced target parent. Stale locks fail closed and are never reclaimed from host-local PID evidence. The stricter Python helper is deliberately bypassed only in this mode because it requires rename to preserve inode identity.

## Implementation

- Export `RenameIdentityPolicy` through the root API.
- Add SHA-256 verification when the source temp fd and renamed destination report different identities.
- Preserve strict defaults, `already-exists`, `mkdir: false`, stale-lock, and post-write identity behavior.
- Document the compatibility boundary and add an `0.4.1 - Unreleased` changelog entry thanking @jlautman.
- Add deterministic regressions for strict/compatible identity behavior, content replacement, per-call overrides, required-Python routing, create semantics, missing parents, stale locks, and normal POSIX cleanup.

## Validation

- `pnpm check`: 38 test files passed; 423 tests passed, 3 skipped.
- `pnpm audit --prod --json`: 0 vulnerabilities.
- Autoreview: no accepted/actionable findings after fixing the no-`mkdir` contract and lock-parent boundary findings.
- Live packaged-consumer proof on AWS Crabbox, Ubuntu 26.04, kernel 7.0, Node 24.18.0, pnpm 11.9.0, rclone 1.74.3: [run `run_b2ca73c0818e`](https://crabbox.openclaw.ai/portal/runs/run_b2ca73c0818e), lease `cbx_dfc891eea183`.
  - rclone rename identity changed from `55:13142417776620928161` to `55:13160412383924634112`.
  - Strict mode returned `path-mismatch`.
  - Compatibility mode wrote and read `compatible-content`.
  - Root-level compatibility lock was absent after completion.

Exact validated head: `46e6cc26cba0f4f2b6e1ea5217cb35a6d34e1bd9`.


steipete:
@clawsweeper re-review

The branch now deliberately routes the opt-in around the Python helper, fails closed on stale locks, preserves `already-exists` and `mkdir: false`, corrects the lock timing docs, keeps per-target locks inside the canonical root, and links exact rclone FUSE proof in the PR body. Please review exact head `46e6cc26cba0f4f2b6e1ea5217cb35a6d34e1bd9`.

steipete:
Landed via rebase on `main` as `09614f41d3a2d0467a7227da827e5cae8b76cdb8` through `7c2cfa1b6517153be141d1771bf81dc231f73bec`.

Proof:

- Local `pnpm check`: 38 test files passed; 423 tests passed, 3 skipped.
- Local `pnpm audit --prod --json`: 0 vulnerabilities.
- Autoreview completed with no accepted/actionable findings after the lock-parent and `mkdir: false` regressions were fixed.
- Live packaged-consumer rclone FUSE proof: [AWS Crabbox run `run_b2ca73c0818e`](https://crabbox.openclaw.ai/portal/runs/run_b2ca73c0818e), lease `cbx_dfc891eea183`. Strict mode returned `path-mismatch`; the explicit compatibility mode wrote and read the expected content; its root-level lock was removed.
- Exact-head GitHub checks passed: [CI](https://github.com/openclaw/fs-safe/actions/runs/28498361204), [coverage](https://github.com/openclaw/fs-safe/actions/runs/28498361190), and [benchmarks](https://github.com/openclaw/fs-safe/actions/runs/28498361199).

Caveat: `verify-content-with-lock` is intentionally weaker than strict object-identity verification and is limited to trusted/cooperating writers; strict identity remains the default. Thanks @jlautman for the report and contribution.

jlautman:
Thanks for the improvements since we opened this. Three things came up in review that you may want to address or dismiss before merging.

**Stream input with `verify-content-with-lock` throws on FUSE**

In `runPinnedWriteFallback`, when inode churn fires and the content-hash branch runs, a stream input hits:

```ts
if (params.input.kind !== "buffer") {
  throw new FsSafeError("path-mismatch", "fallback target changed during write");
}
```

The stream was consumed writing to the temp file, so there's nothing to hash. But the write succeeded, and the lock means no cooperating writer could have interleaved. Throwing `path-mismatch` with "changed during write" is wrong here: the file is fine, the error is misleading, and a caller that retries on `path-mismatch` will attempt a double-write with an already-consumed stream. A safer behavior would be to open the target by fd, take the fd-based stat as the stable identity, and return without content verification, since the lock already provides the mutual exclusion that makes that safe.

**`O_NOFOLLOW` flag computed twice in the same function**

`runPinnedWriteFallback` has the same three-line platform check inline twice: once for the temp file open flags and once for the read handle. Other files in the repo define a module-level `SUPPORTS_NOFOLLOW` constant instead. A single constant would remove the duplication.

**Dead `"throw"` variant in `RenameIdentityMismatchPolicy`**

```ts
type RenameIdentityMismatchPolicy = "throw" | "verify-content";
```

`"throw"` is never passed at any call site. The runtime check (`!== "verify-content"`) treats it identically to `undefined`. Narrowing the type to just `"verify-content"` removes the implication that the two variants do different things.

---

Happy to send a patch for any of these, but didn't want to push more commits without checking first.

PR #39, opened 2026-07-13
fix(guarded-mkdir): allow in-root symlinked directory components

ctbritt:
## Summary

`mkdirPathComponentsWithGuards` rejected any existing path component that was a symlink, even when it resolved to a real directory inside the root — throwing `"directory component must be a directory"` deterministically (not intermittently), regardless of any cache state.

This breaks any pinned write whose path traverses a symlinked directory that legitimately lives inside the boundary — for example a "skill-bank" layout where `<root>/skills/<name>` is a symlink into `<root>/skills-bank/skills/<name>`. Found while debugging [openclaw/openclaw#106693](https://github.com/openclaw/openclaw/issues/106693).

The function already realpath-checks each segment against the root immediately after the offending check (per its own comment: *"Build one segment at a time and realpath-check each segment before descending"*), so the safety check the original author intended was simply placed after an unconditional throw instead of gating it — the throw fires before the realpath/in-root check ever gets a chance to run.

## Fix

Only throw `not-file` when a component is neither a symlink nor a directory. For an existing symlink:
1. Resolve it and verify (as the code already did) that the resolved real path is still inside the root — reject with `outside-workspace` exactly as before if not.
2. Confirm the resolved target is itself a directory — reject with `not-file` if not.
3. Continue the walk using the **resolved real path**, so every subsequent guard check in the loop operates on a real (non-symlink) directory, preserving TOCTOU protection for the rest of the walk.

No existing security invariant is weakened: a symlink escaping the root is still rejected, a symlink pointing at a non-directory is still rejected, and a plain-file component is still rejected.

## Test plan

- [x] New dedicated test file `test/guarded-mkdir-symlink-parent.test.ts`, 4 cases:
  - in-root symlinked directory component now succeeds (the fix)
  - symlink escaping the root still rejected (`outside-workspace`)
  - plain file component still rejected (`not-file`)
  - symlink pointing at a file (not a directory) still rejected (`not-file`)
- [x] `pnpm check` (file-size lint + fs-boundary lint + build + full test suite) — all green, 425 passed / 3 pre-existing skips
- [x] `pnpm test:security` (boundary-bypass + adversarial-payload suites) — all green, 59/59

🤖 Generated with assistance from [Claude Code](https://claude.com/claude-code), reviewed and directed by @ctbritt.

### DIFF
--- a/src/guarded-mkdir.ts
+++ b/src/guarded-mkdir.ts
@@ -35,13 +35,30 @@ export async function mkdirPathComponentsWithGuards(params: {
       }
     }
     const stat = await fs.lstat(next);
-    if (stat.isSymbolicLink() || !stat.isDirectory()) {
+    if (!stat.isSymbolicLink() && !stat.isDirectory()) {
       throw new FsSafeError("not-file", "directory component must be a directory");
     }
     // Node's recursive mkdir follows symlinks in missing components. Build one
     // segment at a time and realpath-check each segment before descending.
-    if (!isSameOrChildPath(path.resolve(await fs.realpath(next)), rootCanonical)) {
+    const nextReal = path.resolve(await fs.realpath(next));
+    if (!isSameOrChildPath(nextReal, rootCanonical)) {
       throw new FsSafeError("outside-workspace", "directory escaped workspace root");
+    }
+    if (stat.isSymbolicLink()) {
+      // An existing path component may legitimately be a symlink to a real
+      // directory inside the root (e.g. a skill-bank layout). We already
+      // verified above that it resolves inside the root, so treat the
+      // resolved real path as the directory for the rest of this walk
+      // instead of rejecting it outright. Guard checks from here on operate
+      // on the real (non-symlink) path, preserving TOCTOU protection for
+      // every subsequent segment.
+      const targetStat = await fs.stat(nextReal);
+      if (!targetStat.isDirectory()) {
+        throw new FsSafeError("not-file", "directory component must be a directory");
+      }
+      await createAsyncDirectoryGuard(nextReal);
+      current = nextReal;
+      continue;
     }
     await createAsyncDirectoryGuard(next);
     await assertAsyncDirectoryGuard(parentGuard);
--- a/src/guarded-mkdir.ts
+++ b/src/guarded-mkdir.ts
@@ -2,18 +2,39 @@ import fs from "node:fs/promises";
 import path from "node:path";
 import { assertAsyncDirectoryGuard, createAsyncDirectoryGuard } from "./directory-guard.js";
 import { FsSafeError } from "./errors.js";
-import { isPathRelativeEscape } from "./path.js";
+import { isNotFoundPathError, isPathRelativeEscape } from "./path.js";
 
 function isSameOrChildPath(candidate: string, parent: string): boolean {
   const parentPrefix = parent.endsWith(path.sep) ? parent : `${parent}${path.sep}`;
   return candidate === parent || candidate.startsWith(parentPrefix);
 }
 
+async function realpathOrThrowNotFile(target: string): Promise<string> {
+  try {
+    return path.resolve(await fs.realpath(target));
+  } catch (error) {
+    if (isNotFoundPathError(error)) {
+      // A dangling symlink (or a component removed between lstat and
+      // realpath) is not a usable directory component.
+      throw new FsSafeError("not-file", "directory component must be a directory", {
+        cause: error instanceof Error ? error : undefined,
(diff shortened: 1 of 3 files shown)


ctbritt:
Both findings were correct, thanks — pushed 0c4ede0 to address them:

1. **Resolved parent path not returned to fallback writers** — you're right, my original unit test only exercised \`mkdirPathComponentsWithGuards\` in isolation and never actually proved the end-to-end write worked, so this gap slipped through. \`mkdirPathComponentsWithGuards\` now returns the resolved real path of the final component; \`runPinnedWriteWithRenamePolicy\` reassigns its local \`parentPath\` to that returned value when \`mkdir: true\`, so the subsequent \`createAsyncDirectoryGuard\` call and the constructed \`targetPath\` both use the real (non-symlink) directory instead of re-deriving from the original lexical path. Added a real end-to-end test through \`root().write()\` (forcing the JS fallback path via \`configureFsSafePython({ mode: "off" })\`) that writes directly into a symlinked skill-bank-style directory and confirms the content lands at the real target — this is the actual scenario that motivated the fix, and it wasn't covered before.

2. **Dangling symlink realpath failure unwrapped** — fixed. The \`fs.realpath()\` call for an existing symlink component now catches ENOENT and throws a typed \`FsSafeError("not-file", ...)\` instead of letting the raw Node error propagate. Added a dedicated regression test.

Full suite (427/430, 3 pre-existing skips) and the security/boundary-bypass suite (59/59) both still pass.

One related-but-separate thing I noticed while fixing this and am leaving out of scope for this PR: \`createNearestExistingDirectoryGuard\` (used on the \`mkdir: false\` path) has the same unconditional symlink rejection via its own calls to \`createAsyncDirectoryGuard\`, so an in-root symlinked parent would still be rejected on a write where the directory already fully exists and \`mkdir\` isn't requested. Didn't want to widen this PR further, but flagging it in case it's worth a follow-up.

ctbritt:
Good catch on the Python-helper/Node-fallback inconsistency — that's a real gap and I don't want to paper over it. A couple of things worth surfacing for the maintainer decision you're asking for:

**This PR's motivating bug only ever hits the Node fallback path.** OpenClaw (the actual downstream consumer that surfaced this) explicitly disables the Python helper by default — its own bundled init code calls \`configureFsSafePython({ mode: "off" })\` unless \`FS_SAFE_PYTHON_MODE\`/\`OPENCLAW_FS_SAFE_PYTHON_MODE\` is set, which it isn't in the environment that reported the original bug. So for the concrete case that motivated this fix (skill-bank layout, `mkdir: true`, no Python helper), this PR is correct and complete as-is — verified end-to-end in the test suite.

**I don't want to guess my way through Python-helper parity.** Making `runPinnedWriteHelper`'s `os.open(..., O_NOFOLLOW)` walk behave consistently with the Node fallback is a real, separate change to a different-language implementation of the same security-sensitive contract, and it deserves its own careful review rather than being bolted onto this PR by someone without the full context on why the Python side chose `O_NOFOLLOW` in the first place (rename-safety? a different threat model? just not updated yet?). That's exactly the kind of call the review flagged as needing maintainer sign-off, not a contributor guess.

**Proposing option 2 from the review:** merge this as an explicitly fallback-only fix, with a note that helper-enabled (`auto`/`require`) mode still rejects in-root symlinked parents pending a follow-up that gives the Python helper equivalent containment-preserving traversal. I'll open a tracking issue for that follow-up if useful, but would rather leave the actual Python-side implementation to whoever owns that code's threat model — happy to take a pass at it too if a maintainer confirms the intended contract first.

@steipete — given you're flagged as the strongest ownership trail on this behavior, any guidance on whether helper/fallback parity is expected here, or whether fallback-only is an acceptable interim contract?

steipete:
Fallback-only is an acceptable interim contract — merging on that basis. The Python helper's `O_NOFOLLOW` rejection stays as-is for now: it fails closed, and its threat model shouldn't be reworked as a rider on this fix.

Two follow-ups, both maintainer-side so nothing blocks this PR:

1. The symlink branch skips the final `assertAsyncDirectoryGuard(parentGuard)` re-assert. Since the walk continues on the frozen resolved real path, containment doesn't depend on it — but we'll add a comment (or the assert) in the 0.5.0 branch so the asymmetry is documented.
2. Helper parity: the plan for 0.5.0 is to replace the Python helper with a small native addon using `openat2(RESOLVE_BENEATH)` on Linux, which gives in-root symlink traversal with kernel-level containment — making both paths consistent by construction. Tracking issue to follow.

Thanks for the careful fix and for not guessing your way through the Python side.

PR #67, opened 2026-07-27
fix(path): reject Windows reserved basenames in sanitizeUntrustedFileName

SebTardif:
## What Problem This Solves

Fixes an issue where consumers using `sanitizeUntrustedFileName()` (via staging, external output, and sibling-temp helpers) would accept Windows reserved device basenames such as `CON`, `NUL.txt`, or `COM1` from untrusted input. On Windows those names open devices rather than ordinary files, which breaks portable name sanitization already used for C0/C1 and Windows-invalid characters on every host.

## Why This Change Was Made

Reuse the reserved-name set and normalization already used by the device-path read guards (`isWindowsReservedDeviceBaseName`) inside `sanitizeUntrustedFileName()`. Reserved basenames (including extensions and trailing spaces/dots) return the caller-supplied fallback. Non-goals: changing path resolution or read guards; only the untrusted basename sanitizer.

## User Impact

Callers that pass reserved basenames now receive the same fallback they already get for empty, `.`, or `..` names. Valid names such as `console.txt` and `null.pdf` are unchanged. No public API signature changes.

### Red / green

- **Red:** `pnpm exec vitest run test/filename.test.ts` failed with `expected 'CON' to be 'fallback.bin'` before the production change.
- **Green:** same file, 4/4 tests pass after the fix; full `pnpm check` (598 passed, 22 skipped).

### Runtime sample (built `dist/`)

```console
"CON" -> fallback.bin
"nul.txt" -> fallback.bin
"COM1" -> fallback.bin
"LPT9.doc" -> fallback.bin
"aux " -> fallback.bin
"console.txt" -> console.txt
"report.pdf" -> report.pdf
```

### Checklist

### Related

- Same-repo reserved-name read guards: [#22](https://github.com/openclaw/fs-safe/pull/22)
- Hardened staging / portable basename sanitization: [#62](https://github.com/openclaw/fs-safe/pull/62)
- Node.js path reserved-name edge cases: [nodejs/node#64266](https://github.com/nodejs/node/pull/64266)

## Real behavior proof

- **Behavior or issue addressed:** Untrusted basenames that are Windows reserved devices must not pass through `sanitizeUntrustedFileName`; they must fall back.
- **Real environment tested:** macOS arm64, Node from repo toolchain, `@openclaw/fs-safe` branch `fix/windows-reserved-basenames` after `pnpm build`.
- **Exact steps or command run after this patch:**

  ```bash
  pnpm install --frozen-lockfile
  pnpm exec vitest run test/filename.test.ts
  pnpm build
  node --input-type=module -e "import { sanitizeUntrustedFileName } from './dist/filename.js';
  for (const c of ['CON','nul.txt','console.txt']) console.log(c, '->', sanitizeUntrustedFileName(c, 'fallback.bin'));"
  pnpm check
  ```

- **Evidence after fix:** reserved cases map to `fallback.bin`; non-reserved `console.txt` kept; full check green.
- **Observed result after fix:** `sanitizeUntrustedFileName('CON', 'fallback.bin') === 'fallback.bin'` and siblings; staging helpers inherit the safer basename policy.
- **What was not tested:** Live Windows NTFS create of reserved names (behavior is documented OS semantics; this PR blocks the name before open).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Security and Correctness
+
+- **Security — `sanitizeUntrustedFileName()`:** Reject Windows reserved device basenames (`CON`, `PRN`, `AUX`, `NUL`, `COM1`–`COM9`, `LPT1`–`LPT9`, and variants with extensions or trailing spaces/dots) and return the caller fallback, matching the reserved-name rules already used by device-path read guards. Prevents staging/output helpers from accepting untrusted names that open devices on Windows.
+
 ## 0.5.0 - 2026-07-26
 
 ### Highlights
--- a/src/device-path.ts
+++ b/src/device-path.ts
@@ -104,6 +104,11 @@ function normalizeWindowsDeviceBaseName(filePath: string): string {
   return (withoutTrailingIgnoredChars.split(".")[0] ?? withoutTrailingIgnoredChars).toUpperCase();
 }
 
+/** True when the basename is a Windows reserved device name (CON, NUL, COM1, …). */
+export function isWindowsReservedDeviceBaseName(fileName: string): boolean {
+  return WINDOWS_RESERVED_DEVICE_NAMES.has(normalizeWindowsDeviceBaseName(fileName));
+}
+
 function matchWindowsDeviceReadPath(filePath: string): UnsafeDeviceReadPathMatch | undefined {
   const normalized = filePath.replace(/\//g, "\\");
   if (/^\\\\\.\\/.test(normalized) || /^\\\\\?\\GLOBALROOT\\Device\\/i.test(normalized)) {
--- a/src/filename.ts
+++ b/src/filename.ts
@@ -1,4 +1,5 @@
 import path from "node:path";
+import { isWindowsReservedDeviceBaseName } from "./device-path.js";
 
 const WINDOWS_INVALID_FILE_NAME_CHARACTERS = new Set('<>:"/\\|?*');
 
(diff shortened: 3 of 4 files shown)


steipete:
Maintainer-side reproduction supports CLOSE for the current proposal.

On current `main` (`ab933820c089a6fc6991ed4d80ce0a931ce70567`), `docs/filename.md` explicitly defines the released contract: `sanitizeUntrustedFileName()` deliberately does not reject Windows reserved names, and callers that need that stricter policy layer it on top. After a real build, the exported behavior is:

```text
CON -> CON
nul.txt -> nul.txt
console.txt -> console.txt
```

On this PR's exact head (`c47b4e59718f1eacd2b26645a3445d02c80b372d`), the focused filename test passes (4/4) and the JavaScript emitted by the attempted branch build demonstrates the intended reversal:

```text
CON -> fallback.bin
nul.txt -> fallback.bin
console.txt -> console.txt
```

That is not a bug repair within the existing contract; it is a global compatibility-policy change for every sanitizer and staging/output caller. The existing documented caller-side opt-in is the appropriate boundary unless a separately approved API/default redesign is undertaken.

The branch is also stale/dirty against current `main`. A frozen offline install on this head reverted to its older nine-workspace graph and `pnpm build` failed at `src/archive-tar-runtime.ts` because the `tar` module was unavailable, so the exact head is not a current buildable candidate.

No contributor revision is requested: the recommendation is to close this implementation and retain the documented default.

steipete:
Closing because the proposed default conflicts with the documented `sanitizeUntrustedFileName` contract: Windows reserved-basename policy remains an explicit caller policy rather than a silent universal sanitizer default. Full reproduction and contract rationale: https://github.com/openclaw/fs-safe/pull/67#issuecomment-5154358821

steipete:
The contract discussion here led to an approved redesign. It has now been implemented fresh on top of current `main`—not by reviving this branch—with underscore suffixing that preserves case and extensions (`CON` → `CON_`, `nul.txt` → `nul_.txt`) and uses the complete shared device-name set on every platform.

Commit: https://github.com/openclaw/fs-safe/commit/5ca756d26b080bacf0f27311d9db55a96d1501da

Review PR: https://github.com/openclaw/fs-safe/pull/83

Thanks @SebTardif for prompting the contract discussion and the original portability idea.

PR #75, opened 2026-07-28
fix(path): keep sanitizers responsive on adversarial input

vincentkoc:
## What Problem This Solves

Fixes an issue where consumers passing long attacker-controlled path segments, temp-name inputs, or Windows device paths could trigger excessive CPU use in sanitizer regexes.

## Why This Change Was Made

Backtracking-prone edge and suffix regexes are replaced with bounded character scans while preserving existing path validation, fallback behavior, and Windows device detection. A dedicated regression suite covers the flagged input shapes.

## User Impact

Path and temp-name sanitization remains responsive for very long hostile inputs without changing the public API or accepted output shapes.

## Evidence

- focused test run: 61 tests passed
- `pnpm build`
- file-size lint
- `git diff --check`
- autoreview: clean, no accepted/actionable findings

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Security and Correctness
+
+- Replace backtracking-prone path-segment, temp-name, and Windows device-path sanitizers with linear scans to keep attacker-controlled inputs from causing excessive CPU use.
+
 ### Docs and Tooling
 
 - Publish only the validated release tarball after asserting its byte identity against the release manifest, and tolerate npm registry propagation with bounded exponential backoff.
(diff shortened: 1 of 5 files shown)


vincentkoc:
Maintainer review completed against the prior regex semantics. I ran a deterministic differential check over 50,000 randomized inputs for each changed boundary: path-segment sanitization (both dot-prefix modes), temporary prefix/extension output, and Windows device classification, plus a curated reserved-device corpus. Old and new outputs matched in every case. Existing CI is green across Node 22/24, macOS/Linux/Windows, native, package smoke, coverage, clippy/audit, and workflow lint.

PR #78, opened 2026-08-01
fix(path): keep whitespace in Windows comparison paths

Yigtwxx:
## What Problem This Solves

Fixes an issue where consumers relying on root confinement would have an outside path reported as inside the root when the root string carries surrounding whitespace. The affected surface is Windows path validation and root confinement: `isPathInside` is the predicate other guards build on, and on `win32` its only normalization step is `normalizeWindowsPathForComparison`.

That function ends by delegating to `normalizeLowercaseStringOrEmpty` (`src/string-coerce.ts:31`), which is a free-text string coercion helper — the same module also normalizes fast-mode flags and thread values. Its chain reaches `normalizeNullableString` (`src/string-coerce.ts:5-11`), which calls `value.trim()`. So a path used for containment math is trimmed before it is lowercased, and leading or trailing whitespace silently disappears from the comparison key.

Whitespace is a legal part of a Windows path component, so trimming merges two genuinely different directories into one comparison key: `C:\root ` and `C:\root` both become `c:\root`.

## Why This Change Was Made

The final step now lowercases in place instead of routing the path through the free-text coercion helper. Separator normalization and extended-length (`\\?\`, `\\?\UNC\`) handling are untouched, so the only behavior that shifts is that surrounding whitespace is preserved. Case folding is deliberately left exactly as it was, since it is the documented comparison semantic and is covered by an existing assertion in `test/api-coverage.test.ts`.

Non-goal, stated explicitly: this change does not alter the Unicode case-folding behavior. `toLowerCase()` is not injective — on a Turkish-language Windows install, `"İstanbul"` folds to a 9-code-point string that never round-trips back, so two distinct NTFS names can still collapse onto one comparison key. Deciding whether comparison should move to an ASCII-only or locale-invariant fold changes behavior for every non-ASCII path, which reads as an owner decision rather than a bug fix. I left it out of this PR and can follow up separately if you want it addressed.

## User Impact

Consumers that derive a root from configuration or environment input — where a stray trailing space is easy to introduce — no longer get a false `true` from `isPathInside` for files that live outside that root. Paths without surrounding whitespace are unaffected, and no export, error shape, or default changes.

## Evidence

Reproduction on Windows 11 with Node 24.15.0, using the real exported functions:

```
BEFORE (src/path.ts @ main ab93382)
  normalizeWindowsPathForComparison("C:\\root ")  -> "c:\\root"
  isPathInside("C:\\root ", "C:\\root\\secret.txt") -> true      <- outside path reported as inside

AFTER (this branch)
  normalizeWindowsPathForComparison("C:\\root ")  -> "c:\\root "
  isPathInside("C:\\root ", "C:\\root\\secret.txt") -> false
  isPathInside("C:\\root",  "C:\\root\\secret.txt") -> true      <- genuine descendants unaffected
```

Regression tests were added to `test/windows-path.test.ts`: one platform-independent case for the normalizer (trailing space and a NBSP), one `win32`-only pair for `isPathInside` covering both the padded root and a genuine descendant, plus an assertion that lowercasing, separator normalization, and extended-length stripping still behave as before.

The new tests are load-bearing. Stashing only `src/path.ts` and re-running the file against the unmodified normalizer:

```
BEFORE (fix reverted, tests kept)
  FAIL  Windows comparison normalization > keeps surrounding whitespace so padded paths stay distinct
  FAIL  Windows containment with padded roots > does not report a sibling directory as inside a space-padded root
        Tests  2 failed | 3 passed (5)

AFTER
        Tests  5 passed (5)
```

### What the other `isPathInside` callers do with this change

The review asks for the intended Windows contract across boundary layers, so I inventoried every
in-package caller rather than reasoning about it:

| call site | root argument | effect of preserving whitespace |
|---|---|---|
| `src/archive-staging.ts:124,221,290,321,368` | `destinationRealDir` / `sourceRootReal` (realpath results) | none in practice; exactness is the fail-closed direction |
| `src/file-store-boundary.ts:64,192` · `src/file-store-prune.ts:54` · `src/file-store.ts:431` | `rootReal` / `scopedRoot.rootWithSep` (realpath results) | same |
| `src/file-store.ts:124` | store `rootDir` | same |
| `src/install-path.ts:67` | `path.resolve(baseDir)` | same |
| `src/deny-mutations.ts:49,96,109,110` | **user-supplied `denyMutations` policy entries** | behavior changes — see below |

A realpath result carries the on-disk spelling, so for every confinement caller the change is
either invisible or the intended fail-closed correction. `deny-mutations` is the one caller that
compares free-text configuration, and there the current Windows behavior is worse than "lenient".

### The deny-list case: Windows applies the policy to the wrong directory

The package already pins the intended contract, on POSIX, with two committed tests —
`test/deny-mutations.test.ts` `preserves trailing whitespace in denied paths` and
`… in denied prefixes`. Both carry `it.skipIf(skipOnWindows)`, which is what the trimming
normalizer forced. Driving those same two scenarios through the public `root()` API on Windows 11
(Node 24.15.0), with only `src/path.ts` swapped between runs:

```
denied prefixes entry: "private "        (both "private " and "private" exist on disk)

  main ab93382     write("private /file.txt") -> ALLOWED        <- the denied directory is writable
                   write("private/file.txt")  -> BLOCKED        <- an undenied sibling is blocked

  this branch      write("private /file.txt") -> BLOCKED
                   write("private/file.txt")  -> ALLOWED
```

So on Windows the deny list is not merely approximate today, it is applied to the wrong directory:
the protected path stays writable while its unprotected sibling is refused. The branch produces the
POSIX behavior the repository already documents.

With that established, the two cases no longer need the Windows skip, and they are load-bearing
there:

```
test/deny-mutations.test.ts on Windows
  this branch    Tests  10 passed | 1 skipped (11)
  main path.ts   Tests   2 failed | 8 passed | 1 skipped (11)   <- exactly the two unskipped cases
```

Full suite on Windows moves from `447 passed | 185 skipped` to `449 passed | 183 skipped` — the two
cases that started running, and nothing else. CI already covers `windows-latest` on Node 22 and 24,
so this is enforced rather than local-only. The remaining `skipIf` in that file (symlink ancestors)
is unrelated and untouched.

Validation performed on this branch:

- `pnpm exec vitest run test/windows-path.test.ts` — 5 passed
- `pnpm exec vitest run` (full suite) — 447 passed, 185 skipped (632); 54 files passed, 6 skipped
- `pnpm lint:file-size` — passed
- `pnpm lint:fs-boundary` — passed
- `pnpm build` (`tsc -p tsconfig.json`) — passed
- `git diff --check` — clean

`CHANGELOG.md` has an entry under `Unreleased` -> `Security and Correctness`.

Related: `openclaw/openclaw#109823`, where this normalizer's lowercasing had to be worked around downstream; that PR documents the same helper being copied to obtain a case-preserving variant.

- [x] Tests cover the change
- [x] `pnpm check` gates run locally
- [x] Docs updated when behavior or API changed (no documented behavior changed; the trimming was undocumented)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Security and Correctness
 
+- Stop trimming surrounding whitespace when normalizing Windows paths for comparison, so a space-padded root no longer compares equal to its unpadded sibling and `isPathInside` keeps reporting outside paths as outside.
 - Replace backtracking-prone path-segment, temp-name, and Windows device-path sanitizers with linear scans to keep attacker-controlled inputs from causing excessive CPU use.
 
 ### Docs and Tooling
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Security and Correctness
 
 - Stop trimming surrounding whitespace when normalizing Windows paths for comparison, so a space-padded root no longer compares equal to its unpadded sibling and `isPathInside` keeps reporting outside paths as outside.
+- Apply `denyMutations` prefixes to the directory they name on Windows. A prefix with trailing whitespace previously compared equal to its unpadded sibling, so the named directory stayed writable while the sibling was blocked; the two `preserves trailing whitespace` deny-mutation cases now run on Windows instead of being skipped.
 - Replace backtracking-prone path-segment, temp-name, and Windows device-path sanitizers with linear scans to keep attacker-controlled inputs from causing excessive CPU use.
 
 ### Docs and Tooling
--- a/src/path.ts
+++ b/src/path.ts
@@ -2,7 +2,6 @@ import fs from "node:fs";
 import path from "node:path";
 
 import { FsSafeError } from "./errors.js";
-import { normalizeLowercaseStringOrEmpty } from "./string-coerce.js";
 
 export {
   assertNoUnsafeDeviceReadPath,
@@ -25,7 +24,9 @@ export function normalizeWindowsPathForComparison(input: string): string {
       normalized = `\\\\${normalized.slice(4)}`;
     }
   }
-  return normalizeLowercaseStringOrEmpty(normalized.replaceAll("/", "\\"));
+  // Lowercase only: this value feeds Windows containment math, so surrounding
+  // whitespace is part of the path and must not be trimmed away.
+  return normalized.replaceAll("/", "\\").toLowerCase();
 }
 
 export function isNodeError(value: unknown): value is NodeJS.ErrnoException {
--- a/test/deny-mutations.test.ts
+++ b/test/deny-mutations.test.ts
@@ -116,7 +116,7 @@ describe("root denyMutations policies", () => {
     ).rejects.toMatchObject({ code: "invalid-path" });
   });
 
-  it.skipIf(skipOnWindows)("preserves trailing whitespace in denied paths", async () => {
+  it("preserves trailing whitespace in denied paths", async () => {
     const rootPath = await tempRoot("fs-safe-deny-space-");
     const deniedName = "secret ";
     const trimmedName = "secret";
@@ -131,7 +131,7 @@ describe("root denyMutations policies", () => {
     await expect(readFile(path.join(rootPath, trimmedName), "utf8")).resolves.toBe("allowed");
   });
 
-  it.skipIf(skipOnWindows)("preserves trailing whitespace in denied prefixes", async () => {
+  it("preserves trailing whitespace in denied prefixes", async () => {
     const rootPath = await tempRoot("fs-safe-deny-prefix-space-");
     const deniedDir = path.join(rootPath, "private ");
     const trimmedDir = path.join(rootPath, "private");
--- a/test/windows-path.test.ts
+++ b/test/windows-path.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, it } from "vitest";
 import { isWindowsNetworkPath } from "../src/local-file-access.js";
+import { isPathInside, normalizeWindowsPathForComparison } from "../src/path.js";
 
 describe("Windows path classification", () => {
   it("distinguishes extended local paths from UNC paths", () => {
@@ -11,5 +12,33 @@ describe("Windows path classification", () => {
     ).toBe(true);
     expect(isWindowsNetworkPath("\\\\?\\Volume{abc}\\secrets\\token", "win32")).toBe(true);
     expect(isWindowsNetworkPath("\\\\.\\pipe\\service", "win32")).toBe(true);
+  });
+});
+
+describe("Windows comparison normalization", () => {
+  it("keeps surrounding whitespace so padded paths stay distinct", () => {
+    expect(normalizeWindowsPathForComparison("C:\\root ")).toBe("c:\\root ");
+    expect(normalizeWindowsPathForComparison("C:\\root\u00a0")).toBe(
+      "c:\\root\u00a0",
+    );
+  });
+
+  it("still lowercases, normalizes separators, and strips the extended-length prefix", () => {
+    expect(normalizeWindowsPathForComparison("\\\\?\\UNC\\Server\\Share\\A/../B")).toContain(
+      "\\\\server\\share",
+    );
+    expect(normalizeWindowsPathForComparison("\\\\?\\C:\\Users/Public")).toBe(
+      "c:\\users\\public",
+    );
+  });
+});
+
+describe.skipIf(process.platform !== "win32")("Windows containment with padded roots", () => {
+  it("does not report a sibling directory as inside a space-padded root", () => {
+    expect(isPathInside("C:\\root ", "C:\\root\\secret.txt")).toBe(false);
+  });
+
+  it("still reports genuine descendants as inside", () => {
+    expect(isPathInside("C:\\root", "C:\\root\\secret.txt")).toBe(true);
   });
 });


Yigtwxx:
Answering the maintainer decision the review flagged, since it is the only thing left on this PR.

I intended **option 1 — confirm fail-closed root semantics**, and the patch is written for that reading: `normalizeWindowsPathForComparison` produces a comparison key for containment math, so the lexical spelling of the root has to survive into the key. A space is a legal path component character on Windows, so `C:\root ` and `C:\root` are two different directories and merging them into one key is what produced the false `true`.

On the compatibility risk, to scope what actually changes:

- Callers whose root has no surrounding whitespace: unchanged.
- Callers who pass a padded root *and* padded candidates: unchanged — both sides keep the same spelling and still compare equal.
- Callers who pass a padded root against unpadded candidates: this is the case that flips, and it flips from "reported inside" to "reported outside". That direction is fail-closed, so the failure mode is a denied operation rather than an escape.

Option 2 (normalize root input at the configuration boundary) is not mutually exclusive with this and I think it is the better long-term shape — but it belongs where roots are constructed, not inside a containment predicate, and it does not fix the predicate for callers who build paths themselves. Happy to open that as a follow-up if you want it.

Also flagged in the PR body and worth repeating here: `toLowerCase()` is not injective, so this key can still collapse distinct NTFS names on non-ASCII input (Turkish `İ` being the obvious one). That is pre-existing, deliberately out of scope here, and a separate follow-up if you want comparison moved to an ASCII-only or locale-invariant fold.

No further changes planned from my side — the branch is at `c0696e5` and I'll hold it there unless you'd like something reshaped.

Yigtwxx:
Both remaining items are about the intended contract across boundary layers, so I measured it instead of arguing it. `bb5b29b` adds what the measurement produced; the PR body carries the full write-up.

**The contract is already yours, and it is pinned by committed tests.** `test/deny-mutations.test.ts` has `preserves trailing whitespace in denied paths` and `… in denied prefixes`. Both carry `it.skipIf(skipOnWindows)` — the skip exists because the trimming normalizer made Windows disagree with them.

**On Windows the deny list is currently applied to the wrong directory.** Driving those two scenarios through the public `root()` API on Windows 11 / Node 24.15.0, swapping only `src/path.ts` between runs, with both `private ` and `private` present on disk:

```
denied prefixes entry: "private "

  main ab93382   write("private /file.txt") -> ALLOWED     <- the denied directory is writable
                 write("private/file.txt")  -> BLOCKED     <- an undenied sibling is blocked

  this branch    write("private /file.txt") -> BLOCKED
                 write("private/file.txt")  -> ALLOWED
```

That is stronger than the `isPathInside` example already in the body: it is not an approximation, it is an inversion, and it affects the mutation guard rather than only the containment predicate.

Given that, I dropped the Windows skip on those two cases. They are load-bearing there:

```
test/deny-mutations.test.ts on Windows
  this branch    Tests  10 passed | 1 skipped (11)
  main path.ts   Tests   2 failed | 8 passed | 1 skipped (11)
```

Full Windows suite: `447 passed | 185 skipped` → `449 passed | 183 skipped`. Nothing else moved. `ci.yml` already runs `windows-latest` on Node 22 and 24, so this is now enforced in CI rather than only on my machine. The third `skipIf` in that file (symlink ancestors) is unrelated and untouched.

**On the compatibility item:** every other in-package caller passes a realpath result (`archive-staging`, `file-store*`, `install-path`), where the on-disk spelling is authoritative and exactness is the fail-closed direction. `deny-mutations` was the only caller reading free-text configuration, and it is the one the change repairs. So the blast radius for "a caller that supplied a padded root and relied on the trim" is limited to callers outside this package; if you would rather keep that leniency for policy input specifically, the coercion belongs at the policy boundary (`policyPathEntries`, next to `safeDirName`'s existing `trim()` in `install-path.ts:7`) rather than in the comparison primitive — happy to add that instead if you prefer it.

CHANGELOG updated with the deny-mutations effect under the existing `Unreleased` entry.

Yigtwxx:
One correction to the note above, so the claim is not read as stronger than it is.

`ci.yml` does define `windows-latest` on Node 22 and 24, but no workflow run has executed on this PR: `ci`, `coverage`, `benchmarks`, and `CodeQL` are all sitting at `action_required` on both this head and the previous one, so they are waiting on a maintainer to approve workflows for the fork. Only `ClawSweeper Dispatch` and the Socket checks have run.

So the Windows numbers above are from my machine (Windows 11, Node 24.15.0), and the unskipped cases will only be enforced once you approve the run. If you would rather not spend a CI approval on it yet, the local before/after is reproducible with:

```
git checkout ab93382 -- src/path.ts && pnpm exec vitest run test/deny-mutations.test.ts   # 2 failed
git checkout HEAD -- src/path.ts     && pnpm exec vitest run test/deny-mutations.test.ts   # 10 passed
```

steipete:
Maintainer-side reproduction on the exact head `bb5b29bb99d127e450d7a33b0aaf25f3fbbc0f33` supports LAND.

Verified locally with Node 24.18.0 and pnpm 10.34.5:

- `pnpm exec vitest run test/windows-path.test.ts test/deny-mutations.test.ts` — 14 passed, 2 platform skips.
- `pnpm check` — lint and build passed; full suite 607 passed, 25 skipped. The final temporary consumer install was interrupted after npm registry requests stalled for five minutes, with no package assertion failure emitted.
- Built-library proof after `pnpm build`: `normalizeWindowsPathForComparison("C:\\root ")` returned `"c:\\root "`, while the unpadded path returned `"c:\\root"`; the comparison keys were distinct.
- `git diff --check` — clean.

Source review agrees with the patch shape: path comparison must preserve legal path characters, and any input convenience trimming belongs at a specific configuration boundary rather than inside the shared confinement predicate. This also makes the existing trailing-whitespace deny-policy contract enforceable on Windows in the fail-closed direction.

Hosted Windows CI is still the remaining independent gate; the contributor's real Windows before/after proof covers the platform behavior meanwhile.

PR #79, opened 2026-08-02
chore(deps): refresh runtime and build dependencies

steipete:
## What changed

- Update `@napi-rs/cli` 3.7.4 → 3.8.1, `@types/node` 26.1.1 → 26.1.2, and Vite 8.1.5 → 8.2.0.
- Update the native crate graph: N-API 3.11.0 → 3.12.0, N-API derive 3.6.0 → 3.6.2, N-API build 2.3.2 → 2.4.0, and ZIP 6.0.0 → 8.6.0, plus compatible transitives.
- Raise the private native workspace's accurate minimum Rust version from 1.85 to 1.88, matching the updated N-API and ZIP crates.
- Advance the SHA-pinned CodeQL v4 action to its current moving-tag commit.

This supersedes #77 with the latest `@napi-rs/cli` patch and the rest of the repository-wide freshness pass.

## Major-version review

ZIP 7 removed feature flags this repository does not use. ZIP 8 removed deprecated `DateTime` methods; the native archive code does not call them. The current read/write APIs used here (`ZipArchive`, `ZipWriter`, and `SimpleFileOptions`) remain available. The target is the latest stable 8.6.0, not the 9.0 prerelease line.

pnpm 11.18.0 was evaluated but intentionally not adopted: a clean target-version install expanded the host graph from 147 packages to a 267-package cross-platform optional-target verification/download pass and could not complete reliably. Keeping pnpm 10.34.5 preserves the proven install contract.

## Local proof

- Node 24.18.0, pnpm 10.34.5.
- `pnpm check` — build/package check passed; 605 tests passed, 23 skipped.
- `pnpm test:security` — 62 passed.
- `pnpm docs:site` — built successfully with Vite 8.2.0.
- `pnpm outdated -r --format json` — no remaining npm package updates.
- `pnpm audit --audit-level high --json` — zero vulnerabilities.
- Cargo 1.97.1: `cargo metadata --manifest-path native/Cargo.toml --no-deps --locked --format-version 1` passed; `cargo update --dry-run` reported zero updates.
- AutoReview — clean, no accepted/actionable findings.
- `git diff --check` — clean.

Real built-library integration exercised the public root and archive boundaries: an in-root write/read returned `safe boundary`, `../escape.txt` failed with `outside-workspace`, and a generated ZIP extracted through the guarded JavaScript path with `archive boundary` intact.

Hosted native builds/tests and bundled-package smoke across Linux, macOS, and Windows remain the final proof for the Rust major.

PR #84, opened 2026-08-02
fix(root): report documented remove() failure codes

Yigtwxx:
## What Problem This Solves

Fixes an issue where consumers calling `Root.remove()` would receive the boundary-violation code `path-alias` ("path is not under root") when the target was simply missing or the directory was simply not empty.

The affected surface is root confinement error reporting. `removePathInRoot()` routes every failure through `normalizePinnedPathError()`, which rewrites any non-`FsSafeError` into `path-alias`. Because `removePathFallback()` reaches the filesystem through plain `fs.lstat` / `fs.rmdir` / `fs.rm`, an `ENOENT` and an `ENOTEMPTY` both surface as boundary violations.

The documentation already specifies the intended behavior in six places:

- `docs/writing.md:130,135` and `docs/writing.md:193,194` — non-empty directories throw `not-empty`, `not-removable` covers other unlink/rmdir failures
- `docs/errors.md:96,100` — same two codes, with `not-found` at `docs/errors.md:98`
- `docs/file-store.md:135` — "non-empty dirs throw `not-empty`"
- `docs/root.md:156` — `not-found` for a missing target

`not-empty` and `not-removable` are declared in the exported `FsSafeErrorCode` union (`src/errors.ts:10,14`) and listed in `docs/types.md:156-157`, but before this change they were constructed nowhere in `src/` and asserted nowhere in `test/`. The documentation was already correct; the implementation was not.

## Why This Change Was Made

A new `normalizeRemovePathError()` in `src/root-errors.ts` classifies the failure at the `remove()` call site: `ENOENT`/`ENOTDIR` to `not-found`, `ENOTEMPTY`/`EEXIST` to `not-empty`, and any other errno to `not-removable`. Anything that is not errno-shaped still falls back to `normalizePinnedPathError()`, so today's `path-alias` behavior is preserved for genuinely unclassifiable throws.

Three boundaries were deliberate:

- **Directory guard signals are untouched.** `createAsyncDirectoryGuard()` and `assertAsyncDirectoryGuard()` already throw `FsSafeError` (`not-file` and `path-mismatch`), so the leading `instanceof FsSafeError` check returns them unchanged. TOCTOU identity drift still reports `path-mismatch` — that is covered by a unit assertion.
- **`normalizePinnedPathError()` itself is unchanged.** `mkdirPathInRoot()` shares it and is out of scope for this PR, and `test/edge-coverage.test.ts` pins its current contract.
- **`EEXIST` maps to `not-empty`** because POSIX permits `rmdir` to report `EEXIST` for a non-empty directory. The package already treats the pair as equivalent in `src/move-path.ts:346` and `src/trash.ts:12`.

Compatibility: error codes are a public compatibility surface, and this is a deliberate behavior change. A consumer branching on `path-alias` around `remove()` will observe the new codes. Consumers written against the documented contract are fixed by this change; consumers written against the previous behavior were treating a routine `ENOTEMPTY` as a containment failure, which is the more dangerous of the two readings. No internal caller depends on the old code — `src/file-store-prune.ts:89,95` is the only in-package consumer and it discards remove errors.

Non-goal: `mkdirPathInRoot()` shares the same normalizer and has its own error-shape questions. Left alone to keep this to one concern; happy to open a follow-up.

## User Impact

`Root.remove()` and `FileStore.remove()` now report the codes their documentation already promised:

```ts
await fs.remove("missing.txt");        // not-found     (was: path-alias)
await fs.remove("snapshots/full-dir"); // not-empty     (was: path-alias)
// permissions, device busy, read-only fs -> not-removable (was: path-alias)
```

Directory identity drift during the remove still reports `path-mismatch`, unchanged. No API, option, default, or export changed; no migration is needed beyond widening a `path-alias` catch if one was written against the previous behavior.

## Evidence

Reproduction on `main` before the fix, using the new regression test:

```
FAIL  test/fs-safe.test.ts > @openclaw/fs-safe > reports documented failure codes
      when remove cannot unlink the target
AssertionError: expected FsSafeError: path is not under root { …(3) } to match
object { code: 'not-found' }
- { "code": "not-found",
+ FsSafeError { "code": "path-alias", }
 ❯ test/fs-safe.test.ts:507:45
```

Regression coverage added:

- `test/fs-safe.test.ts` — real-disk `mkdtemp` root exercising `not-found` for a missing target, `not-found` for a missing parent, and `not-empty` for a populated directory. Not platform-skipped: `removePathInRoot()` has no platform branch, so it runs on all six CI matrix legs.
- `test/edge-coverage.test.ts` — unit assertions over the six errno mappings, the `FsSafeError` pass-through for `path-mismatch`, and the non-errno fallback to `path-alias`.

`not-removable` is covered at the unit level rather than end to end on purpose: forcing `EACCES` needs a `chmod` that is a no-op for root in container CI and is ignored on Windows, and forcing `EBUSY` only works on Windows. A platform-skipped end-to-end case would assert less than the unit test does on every platform.

Validation on Windows 11, Node v22.20.0, pnpm 10.34.5:

```
pnpm check
  lint:file-size    pass
  lint:fs-boundary  pass
  build             pass
  test              Test Files  54 passed | 6 skipped (60)
                    Tests  489 passed | 183 skipped (672)
  check-pack.mjs    exit 0

pnpm test:security  Test Files  5 passed (5)
                    Tests  46 passed | 17 skipped (63)

targeted           Tests  21 passed | 13 skipped (34)
                   (test/fs-safe.test.ts, test/edge-coverage.test.ts;
                    19 passed on main before the two new tests)
```

## Update — `e96eb9c`

The errno mapping wrapped the whole fallback, and the parent-directory guard runs inside it, so a raw guard error such as `ELOOP` became `not-removable` although no deletion was attempted.

The guard stage now sits in `prepareRemoveGuard()` with its own normalizer, and the errno mapping wraps only the `lstat`/`rmdir`/`rm` calls on the target. `removePathInRoot()` is back to `normalizePinnedPathError()`, unchanged from `main`.

`normalizeRemoveGuardError()` keeps `FsSafeError` unchanged, maps `ENOENT`/`ENOTDIR` to `not-found` — the parent does not exist, so the target is definitionally absent and nothing mutated — and fails closed with `path-alias` for every other raw error.

Pre-fix, on head `44cc206` with the new regression:

```
FAIL  test/edge-coverage.test.ts > root error helpers >
      does not map a guard-stage filesystem failure to a removal code
AssertionError: expected FsSafeError: path could not be removed { …(3) } to match
object { code: 'path-alias' }
- Expected
+ Received
- {
-   "code": "path-alias",
+ FsSafeError {
+   "code": "not-removable",
```

Coverage added: an end-to-end case driving an errno-shaped guard failure through the existing `beforeRootFallbackMutation` hook, so it is deterministic on all six CI legs, plus unit assertions over every branch of the new normalizer.

Validation on Windows 11, Node v22.20.0:

```
lint:file-size    pass
lint:fs-boundary  pass
tsc --noEmit      pass
vitest run        Test Files  54 passed | 6 skipped (60)
                  Tests  491 passed | 183 skipped (674)
test:security     Test Files  5 passed (5)
                  Tests  46 passed | 17 skipped (63)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Security and Correctness
 
-- Report the documented `remove()` failure codes instead of collapsing every failure to `path-alias`. A missing target now throws `not-found`, a non-empty directory throws `not-empty`, and any other filesystem failure throws `not-removable`, while directory identity drift keeps reporting `path-mismatch`.
+- Report the documented `remove()` failure codes instead of collapsing every failure to `path-alias`. A missing target now throws `not-found`, a non-empty directory throws `not-empty`, and any other filesystem failure throws `not-removable`, while directory identity drift keeps reporting `path-mismatch`. The errno classification covers only the deletion syscalls, so a parent-directory guard failure still fails closed with `path-alias` rather than implying a removal was attempted.
 
 ## 0.5.1 - 2026-08-01
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Suffix Windows reserved basenames with `_` in `sanitizeUntrustedFileName()` while preserving case and extensions on every platform, including dollar names and superscript COM/LPT variants; thanks @SebTardif (#67).
 
+### Security and Correctness
+
+- Report the documented `remove()` failure codes instead of collapsing every failure to `path-alias`. A missing target now throws `not-found`, a non-empty directory throws `not-empty`, and any other filesystem failure throws `not-removable`, while directory identity drift keeps reporting `path-mismatch`.
+
 ## 0.5.1 - 2026-08-01
 
 ### Security and Correctness
--- a/src/root-errors.ts
+++ b/src/root-errors.ts
@@ -25,6 +25,18 @@ export function normalizePinnedPathError(error: unknown): Error {
   });
 }
 
+export function normalizeRemoveGuardError(error: unknown): Error {
+  if (error instanceof FsSafeError) {
+    return error;
+  }
+  if (isNotFoundPathError(error)) {
+    return new FsSafeError("not-found", "file not found", {
+      cause: error instanceof Error ? error : undefined,
+    });
+  }
+  return normalizePinnedPathError(error);
+}
+
 export function normalizeRemovePathError(error: unknown): Error {
   if (error instanceof FsSafeError) {
     return error;
--- a/src/root-errors.ts
+++ b/src/root-errors.ts
@@ -1,5 +1,7 @@
 import { FsSafeError } from "./errors.js";
-import { hasNodeErrorCode } from "./path.js";
+import { hasNodeErrorCode, isNodeError, isNotFoundPathError } from "./path.js";
+
+const REMOVE_NOT_EMPTY_CODES = new Set(["ENOTEMPTY", "EEXIST"]);
 
 export function isAlreadyExistsError(error: unknown): boolean {
   return hasNodeErrorCode(error, "EEXIST") || /File exists|EEXIST/i.test(String(error));
@@ -21,4 +23,21 @@ export function normalizePinnedPathError(error: unknown): Error {
   return new FsSafeError("path-alias", "path is not under root", {
     cause: error instanceof Error ? error : undefined,
   });
+}
+
+export function normalizeRemovePathError(error: unknown): Error {
+  if (error instanceof FsSafeError) {
+    return error;
+  }
+  if (!isNodeError(error) || typeof error.code !== "string") {
+    return normalizePinnedPathError(error);
+  }
+  const cause = error instanceof Error ? error : undefined;
+  if (isNotFoundPathError(error)) {
+    return new FsSafeError("not-found", "file not found", { cause });
+  }
+  if (REMOVE_NOT_EMPTY_CODES.has(error.code)) {
+    return new FsSafeError("not-empty", "directory is not empty", { cause });
+  }
+  return new FsSafeError("not-removable", "path could not be removed", { cause });
 }
--- a/src/root-impl.ts
+++ b/src/root-impl.ts
@@ -47,6 +47,7 @@ import {
   isAlreadyExistsError,
   normalizePinnedPathError,
   normalizePinnedWriteError,
+  normalizeRemovePathError,
 } from "./root-errors.js";
 import { getFsSafeTestHooks } from "./test-hooks.js";
 import { stringifyJsonDocument } from "./json-stringify.js";
@@ -988,7 +989,7 @@ async function removePathInRoot(
   try {
     await removePathFallback(resolved);
   } catch (error) {
-    throw normalizePinnedPathError(error);
+    throw normalizeRemovePathError(error);
   }
 }
 
--- a/src/root-impl.ts
+++ b/src/root-impl.ts
@@ -47,6 +47,7 @@ import {
   isAlreadyExistsError,
   normalizePinnedPathError,
   normalizePinnedWriteError,
+  normalizeRemoveGuardError,
   normalizeRemovePathError,
 } from "./root-errors.js";
 import { getFsSafeTestHooks } from "./test-hooks.js";
@@ -989,7 +990,7 @@ async function removePathInRoot(
   try {
     await removePathFallback(resolved);
   } catch (error) {
-    throw normalizeRemovePathError(error);
+    throw normalizePinnedPathError(error);
   }
 }
 
@@ -1325,11 +1326,24 @@ async function resolvePinnedRootPathInRoot(
   };
 }
 
+async function prepareRemoveGuard(targetPath: string) {
+  try {
+    const guard = await createAsyncDirectoryGuard(path.dirname(targetPath));
+    await getFsSafeTestHooks()?.beforeRootFallbackMutation?.("remove", targetPath);
+    await assertAsyncDirectoryGuard(guard);
+    return guard;
+  } catch (error) {
+    throw normalizeRemoveGuardError(error);
+  }
+}
+
 async function removePathFallback(resolved: { resolved: string }): Promise<void> {
-  const guard = await createAsyncDirectoryGuard(path.dirname(resolved.resolved));
-  await getFsSafeTestHooks()?.beforeRootFallbackMutation?.("remove", resolved.resolved);
-  await assertAsyncDirectoryGuard(guard);
-  await ((await fs.lstat(resolved.resolved)).isDirectory() ? fs.rmdir(resolved.resolved) : fs.rm(resolved.resolved));
+  const guard = await prepareRemoveGuard(resolved.resolved);
+  try {
+    await ((await fs.lstat(resolved.resolved)).isDirectory() ? fs.rmdir(resolved.resolved) : fs.rm(resolved.resolved));
+  } catch (error) {
+    throw normalizeRemovePathError(error);
+  }
   await assertAsyncDirectoryGuard(guard).catch(() => undefined);
 }
 
--- a/test/edge-coverage.test.ts
+++ b/test/edge-coverage.test.ts
@@ -20,6 +20,7 @@ import {
   isAlreadyExistsError,
   normalizePinnedPathError,
   normalizePinnedWriteError,
+  normalizeRemovePathError,
 } from "../src/root-errors.js";
 import { movePathToTrash } from "../src/trash.js";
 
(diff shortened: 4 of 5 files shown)


Yigtwxx:
The two Windows `check` legs are red. I looked into it before assuming it was mine, and it is a pre-existing flake rather than a regression from this branch.

What failed here:

```
FAIL test/new-primitives.test.ts > file locks
     > supports manager lifecycle and top-level withFileLock
```

**That exact test also fails on `main`.** It went red in the `release: 0.5.1 (#80)` run, alongside `file-lock-reentrancy.test.ts > continues to arbitrate with a separate process`.

`test/fs-safe.test.ts`, which carries this PR's regression test, passed on the same Windows leg (`26 tests | 12 skipped`).

`ci.yml` history on `main`, most recent first:

| run | result | failing test |
| --- | --- | --- |
| #83 `feat(filename)` | failure | `json.test.ts > recovers readJson from a concurrent real atomic rewrite` (ubuntu) |
| #82 `chore(release)` | failure | `new-primitives.test.ts > secure file reads > reads from a validated Windows ACL and owner` |
| #80 `release: 0.5.1` | failure | `new-primitives.test.ts > file locks > supports manager lifecycle and top-level withFileLock` — the same test failing here — plus `file-lock-reentrancy.test.ts > continues to arbitrate with a separate process` |
| #79 `chore(deps)` | failure | `file-lock-reentrancy.test.ts > queues a absent owner so both read-modify-write operations land` |
| #78 | success | last green run on `main` |

A different test fails almost every run, across both Windows and ubuntu, and all of them are concurrency or file-handle timing cases. On my Windows machine `file-lock-reentrancy.test.ts` failed once with `EPERM` on a `.lock` file and then passed 4/4 in isolation, which matches the same pattern.

This change cannot reach that code: `normalizeRemovePathError()` is only called from `removePathInRoot()`, and `src/sidecar-lock.ts` / `src/file-lock.ts` do not import anything from `src/root-errors.ts`. The green legs on this PR (ubuntu 22/24, macOS 22/24, all native checks, CodeQL, package smoke) exercise the new code paths.

I cannot re-run the job myself — that needs repository admin rights — so if you would like a clean run before reviewing, a maintainer re-run or a rebase once `main` is green would both work; happy to do the rebase.

If the flakes are worth chasing separately, I am glad to open an issue with the run links collected above rather than expanding this PR.

Yigtwxx:
Pushed `e96eb9c` for the guard-boundary finding. It is real: the errno mapping was applied to the whole fallback, and the parent-directory guard runs inside it.

## What was wrong

`removePathFallback()` creates the guard, fires the test hook, and asserts the guard *before* it touches the target:

```
createAsyncDirectoryGuard(dirname)   <- lstat + realpath
assertAsyncDirectoryGuard(guard)     <- lstat + realpath
lstat(target) -> rmdir | rm          <- the only deletion
```

`createAsyncDirectoryGuard()` and `assertAsyncDirectoryGuard()` only convert *known* unsafe states into `FsSafeError` (`not-file`, `path-mismatch`). Any other raw errno from those `lstat`/`realpath` calls escaped, and with one outer `normalizeRemovePathError()` it became `not-removable` — reporting a failed deletion when no deletion was attempted, and losing the fail-closed reading of an unresolved identity condition.

## What changed

The guard stage moved into `prepareRemoveGuard()` with its own normalizer, and the errno mapping now wraps only the deletion calls. `removePathInRoot()` is back to `normalizePinnedPathError()`, unchanged from `main`; everything below it already throws `FsSafeError`, so it is a pass-through.

`normalizeRemoveGuardError()`:

| From the guard stage | Code | Why |
|---|---|---|
| `FsSafeError` | unchanged | `not-file`, `path-mismatch` keep their meaning |
| `ENOENT`, `ENOTDIR` | `not-found` | the parent does not exist, so the target is definitionally absent and nothing mutated — this is `docs/root.md:156` |
| any other errno, non-errno | `path-alias` | fail closed, exactly as on `main` |

The `ENOENT`/`ENOTDIR` carve-out is the one place the guard stage still maps to a documented code. It cannot mask an identity condition: identity drift and non-directory components already arrive as `FsSafeError` and return unchanged above that branch. It is also what keeps `remove("missing-dir/missing.txt")` on the documented `not-found` rather than reporting a boundary violation for an ordinary missing parent.

## Evidence

The new end-to-end regression throws an errno-shaped `ELOOP` from the guard stage. On the previous head:

```
FAIL  test/edge-coverage.test.ts > root error helpers >
      does not map a guard-stage filesystem failure to a removal code
AssertionError: expected FsSafeError: path could not be removed { …(3) } to match
object { code: 'path-alias' }
- Expected
+ Received
- {
-   "code": "path-alias",
+ FsSafeError {
+   "code": "not-removable",
```

That is the finding reproduced: a guard-originated raw error surfacing as `not-removable`. After the fix it is `path-alias`, and the test also asserts the target file is still on disk.

Coverage added to `test/edge-coverage.test.ts`:

- the end-to-end case above, driven through the existing `beforeRootFallbackMutation` hook so it is deterministic on all six CI legs rather than needing a real `ELOOP`/`EACCES`, which is unforgeable as root in container CI and on Windows;
- unit assertions over `normalizeRemoveGuardError()` — `FsSafeError` pass-through, `ENOENT`/`ENOTDIR` to `not-found`, and `ELOOP`/`EACCES`/`EBUSY`/`ENOTEMPTY`/`EEXIST` plus a non-errno throw all to `path-alias`. `ENOTEMPTY` and `EEXIST` are in that list on purpose: they are removal outcomes, so from the guard stage they are nonsense and must fail closed rather than be reported as `not-empty`.

`CHANGELOG.md` now states the boundary alongside the code change.

Validation on Windows 11, Node v22.20.0:

```
lint:file-size    pass
lint:fs-boundary  pass
tsc --noEmit      pass
vitest run        Test Files  54 passed | 6 skipped (60)
                  Tests  491 passed | 183 skipped (674)
test:security     Test Files  5 passed (5)
                  Tests  46 passed | 17 skipped (63)
```

The pre-existing Windows lock flakes noted earlier did not reproduce in this run.

Yigtwxx:
The red Windows leg here isn't from this diff — it's the flake family that's been on `main` since #79, and a re-run moved it.

I closed/reopened this PR to retrigger CI (I can't `gh run rerun` without admin). On **identical code**, the failing test changed:

| Run | Failing leg | Failing test |
|---|---|---|
| First | Windows, Node 22 | `move-path-regression.test.ts > publishes a fresh inode when hardlink rejection is enabled` |
| Re-run | Windows, Node **24** | `file-lock-reentrancy.test.ts > queues a absent owner so both read-modify-write operations land` |

Each time the *other* Windows Node version passed. A real defect fails the same leg every run; this one rotates.

It's also outside this PR's call graph. The diff touches `src/root-errors.ts` and `src/root-impl.ts` (`remove()` errno mapping only). `movePathWithCopyFallback` lives in `src/move-path.ts`, and `file-lock-reentrancy` doesn't reach either changed file.

For context on the first failure specifically — I did chase it before the re-run, in case it was a genuine identity bug. It isn't: with `sourceHardlinks: "reject"` the copy path stages `.fs-safe-move-*.tmp` **while the source is still alive**, so the staged file can't be handed the source's file ID, and `copyEntryWithManifest` bottoms out in `copyRegularFilePinned` with no `link`/clone path. I also ran the exact scenario 200× on a Windows box: 0 collisions. So the `ino` assertion looks sound and the failure is timing/environment, not semantics.

The `dispatch` job failure is unrelated infra — the ClawSweeper bot installation hit its own GitHub API rate limit (`installation ID 122230863`, HTTP 403), triggered by the `pull_request_target` hook on my reopen.

Happy to rebase once `main` is green if you'd prefer a clean run before merging.

Yigtwxx:
Correction to my note above: I called the lock failures timing/environment. I chased them properly afterwards and they are a real, fixable defect — #87 has the root cause and the fix.

Windows denies access to a lock file while a just-unlinked directory entry is still being torn down, so a contended `acquire()` gets `EPERM` (errno `-4048`) on a name that is *already gone*. Instrumented at the moment of failure:

```
EPERM-DIAG {"attempt":2,"lstat":"ENOENT","siblings":["state.json"],"retryAfterMs":0}
```

`lstat` says `ENOENT` and a zero-delay retry opens the file. `acquire()` only treated `EEXIST` as contention, so the `EPERM` escaped from the exclusive create and from `readSidecarLockSnapshot()`, which maps only `ENOENT` to a vanished lock.

Two things I had wrong:

- **It predates #79.** `src/sidecar-lock.ts` last changed in #65; the dependency refresh only shifted timing enough to expose it.
- **The rotation was a symptom of a real race, not proof of an environment problem.** It rotates because the window is timing-dependent, not because nothing is broken.

Measured on Windows 11 / Node 22 over `file-lock-reentrancy` + `new-primitives`: 8 failures in 85 runs before, 0 in 110 after.

That covers the `file-lock-reentrancy` and `new-primitives > file locks` legs. The `move-path-regression > publishes a fresh inode` failure is a different symptom and still open — my analysis of that one in the comment above still stands as far as I took it.

This PR's own diff is unrelated to any of it, so #87 is independent of #84.

steipete:
Landed as 6af7f93 via #93 — thank you, and your commits are credited on the squash.

Your diagnosis and your second push were both right. `removePathInRoot()` collapsed every non-`FsSafeError` into `path-alias`, so the three documented codes were declared in the exported union, documented in six places, and constructed nowhere. Separating the guard stage in `prepareRemoveGuard()` so a raw `ELOOP` is not reported as a removal outcome was a genuinely good catch — that is the kind of thing that usually only turns up after the fix ships.

I added one thing before landing. `src/errors.ts` defines `OPERATIONAL_CODES`, and `not-found`, `not-empty` and `not-removable` were absent from it, so `categorizeFsSafeError()` was still giving all three `category: "policy"` — the category consumers read as "a safety policy rejected this". Since the whole point of the change is that a routine `ENOTEMPTY` is not a boundary violation, fixing the code without the category would only have got you half the way there. All three are now operational, and `docs/types.md` states the rule.

Worth knowing for the downstream bump: this breaks `src/infra/fs-safe-remove.test.ts:66-74` in `openclaw/openclaw`, which asserts `/ENOTEMPTY|EEXIST|EPERM/`; it makes `findPathAliasFilesystemCause()` dead code, since that helper existed purely to unwrap the errno from the bogus `path-alias`; and it quietly fixes three latent bugs there where `not-found` and `not-removable` checks were unreachable.

The red Windows legs on this PR were never yours — that was the sidecar-lock teardown race, fixed separately in #92.

PR #87, opened 2026-08-02
fix(lock): retry Windows lock files denied mid-teardown

Yigtwxx:
## What Problem This Solves

Fixes an issue where consumers taking a contended file lock on Windows would see `acquireFileLock()` reject with `EPERM` when the previous holder released the lock at the same moment. The affected surface is the sidecar file lock (`acquireFileLock`, `withFileLock`, `createFileLockManager`).

Windows denies access to a lock file while a just-unlinked directory entry is still being torn down. In that window the name is already gone but a fresh exclusive create is refused with `ERROR_ACCESS_DENIED`, which libuv reports as `EPERM` (errno `-4048`). The next attempt succeeds, so this is contention, not a permission failure.

`acquire()` treated only `EEXIST` as contention, so the transient `EPERM` escaped from two touch points:

1. the exclusive create, `fs.open(lockPath, "wx")`
2. `readSidecarLockSnapshot()`, which maps only `ENOENT` to a vanished lock

Instrumenting the failure showed the state directly — at the moment of the `EPERM`, the lock file is already gone and a zero-delay retry opens it:

```
EPERM-DIAG {"attempt":2,"lstat":"ENOENT","siblings":["state.json"],"retryAfterMs":0}
```

## Why This Change Was Made

Both touch points now classify a Windows `EPERM` on the lock file as contention and re-enter the existing retry loop. The retry budget is bounded (8 denials), so a genuine permission denial still surfaces as `EPERM` instead of degrading into a `file_lock_timeout` and losing the diagnosis.

The classification is gated to `win32`, matching the equivalence this repository already applies elsewhere — `isPermissionRenameError` in `src/replace-file.ts`, `renameJsonFileWithFallback` in `src/json.ts`, and `windows-rename-denied` in `src/move-path.ts`. `src/sidecar-lock.ts` was the outlier. No public API, error shape, or POSIX behaviour changes.

`src/sidecar-lock.ts` sat at 498 lines against the default 500-line budget, so the predicate lives in `src/sidecar-lock-policy.ts` and the file gets a `LINE_BUDGETS` entry. Happy to split the module instead if you would rather not raise the budget.

## User Impact

Concurrent lock acquisition on Windows no longer fails intermittently when one holder releases while another is waiting. Consumers on POSIX are unaffected. Nothing is added to or removed from the public API.

## Evidence

Two regression tests inject the transient denial at each touch point and assert the acquire still completes. Both fail on `main` and pass with this change:

```
FAIL  test/sidecar-lock-regression.test.ts > retries an exclusive create denied mid-teardown
FAIL  test/sidecar-lock-regression.test.ts > retries a contended snapshot read denied mid-teardown
Tests  2 failed | 11 passed (13)
```

Repeated runs of the real suites on Windows 11 (Node 22), `test/file-lock-reentrancy.test.ts` plus `test/new-primitives.test.ts`:

| | runs | failures |
|---|---|---|
| before | 85 | 8 (~9%) |
| after | 110 | 0 |

The rate before the fix moves with machine load, so a single green run does not demonstrate anything here — the loop does.

Full local verification: `vitest run` 489 passed / 183 skipped, `tsc --noEmit` clean, `scripts/check-file-size.mjs` and `scripts/check-fs-boundary-primitives.mjs` both exit 0.

### Note on the flake reports

This is the defect behind the `file-lock-reentrancy` and `new-primitives > file locks` failures that have been rotating across Windows legs. `src/sidecar-lock.ts` last changed in #65, so it predates the #79 dependency refresh — that bump only shifted timing enough to expose an existing race.

It does not explain `move-path-regression.test.ts > publishes a fresh inode when hardlink rejection is enabled`, which is a different symptom and is still open.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Security and Correctness
+
+- Retry a contended file-lock acquisition when Windows denies access to a lock file whose directory entry is still being torn down. Both the exclusive create and the holder's snapshot read reported that transient `EPERM` as a hard failure, so concurrent `acquireFileLock()` calls failed intermittently on Windows even though the very next attempt would have succeeded. Retries stay bounded, so a genuine permission denial still surfaces as `EPERM` rather than a lock timeout.
+
 ### Features
 
 - Suffix Windows reserved basenames with `_` in `sanitizeUntrustedFileName()` while preserving case and extensions on every platform, including dollar names and superscript COM/LPT variants; thanks @SebTardif (#67).
--- a/scripts/check-file-size.mjs
+++ b/scripts/check-file-size.mjs
@@ -7,9 +7,10 @@ const LINE_BUDGETS = new Map([
   ["src/permissions.ts", 566],
   ["src/root-impl.ts", 1750],
   ["src/root-path.ts", 862],
-  ["src/sidecar-lock.ts", 525],
+  ["src/sidecar-lock.ts", 540],
   ["test/api-coverage.test.ts", 983],
   ["test/new-primitives.test.ts", 1500],
+  ["test/sidecar-lock-regression.test.ts", 540],
 ]);
 
 function walk(dir) {
--- a/scripts/check-file-size.mjs
+++ b/scripts/check-file-size.mjs
@@ -7,7 +7,7 @@ const LINE_BUDGETS = new Map([
   ["src/permissions.ts", 566],
   ["src/root-impl.ts", 1750],
   ["src/root-path.ts", 862],
-  ["src/sidecar-lock.ts", 520],
+  ["src/sidecar-lock.ts", 525],
   ["test/api-coverage.test.ts", 983],
   ["test/new-primitives.test.ts", 1500],
 ]);
--- a/scripts/check-file-size.mjs
+++ b/scripts/check-file-size.mjs
@@ -7,6 +7,7 @@ const LINE_BUDGETS = new Map([
   ["src/permissions.ts", 566],
   ["src/root-impl.ts", 1750],
   ["src/root-path.ts", 862],
+  ["src/sidecar-lock.ts", 520],
   ["test/api-coverage.test.ts", 983],
   ["test/new-primitives.test.ts", 1500],
 ]);
--- a/src/sidecar-lock-policy.ts
+++ b/src/sidecar-lock-policy.ts
@@ -14,11 +14,14 @@ export function computeSidecarLockDelayMs(retry: SidecarLockRetryOptions, attemp
 // is still being torn down, so a contended acquire sees EPERM on a name that is
 // already gone -- both when creating it exclusively and when reading the
 // holder's snapshot. The next attempt succeeds, so this is contention rather
-// than a permission failure.
+// than a permission failure. The error must name the lock file itself: the
+// exclusive-create helper opens the parent directory first, and a denial from
+// that setup step carries no teardown evidence and has to reach the caller.
 export const maxTransientLockDenials = 8;
 
-export function isTransientLockFileDenial(error: unknown): boolean {
-  return process.platform === "win32" && (error as NodeJS.ErrnoException | null)?.code === "EPERM";
+export function isTransientLockFileDenial(error: unknown, lockPath: string): boolean {
+  const denial = error as NodeJS.ErrnoException | null;
+  return process.platform === "win32" && denial?.code === "EPERM" && denial.path === lockPath;
 }
 
 export function sidecarLockPayloadIsStale(
--- a/src/sidecar-lock-policy.ts
+++ b/src/sidecar-lock-policy.ts
@@ -10,6 +10,17 @@ export function computeSidecarLockDelayMs(retry: SidecarLockRetryOptions, attemp
   return Math.min(maxTimeout, Math.round(base * jitter));
 }
 
+// Windows denies access to a lock file while a just-unlinked directory entry
+// is still being torn down, so a contended acquire sees EPERM on a name that is
+// already gone -- both when creating it exclusively and when reading the
+// holder's snapshot. The next attempt succeeds, so this is contention rather
+// than a permission failure.
+export const maxTransientLockDenials = 8;
+
+export function isTransientLockFileDenial(error: unknown): boolean {
+  return process.platform === "win32" && (error as NodeJS.ErrnoException | null)?.code === "EPERM";
+}
+
 export function sidecarLockPayloadIsStale(
   payload: unknown,
   staleMs: number,
--- a/src/sidecar-lock.ts
+++ b/src/sidecar-lock.ts
@@ -266,6 +266,17 @@ export function createSidecarLockManager(key: string) {
     // Bounded so a genuine denial still surfaces as EPERM, not a lock timeout.
     let transientDenials = 0;
     const withinDenialBudget = (): boolean => ++transientDenials <= maxTransientLockDenials;
+    // Waiting can fail on the caller's own retry or deadline limits. Classifying
+    // a denial as contention must not cost them the original diagnosis, so hand
+    // the denial back when no further attempt can be scheduled.
+    const retryOrRethrowDenial = async (denial: unknown): Promise<void> => {
+      try {
+        await waitForRetry();
+      } catch (waitError) {
+        if ((waitError as NodeJS.ErrnoException).code === "file_lock_timeout") throw denial;
+        throw waitError;
+      }
+    };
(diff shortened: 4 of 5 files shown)


Yigtwxx:
CI status on this branch: **22 pass, 1 fail**, and the one red leg is not from this diff.

`Node 22 check (windows-latest)` timed out after 15s in `new-primitives.test.ts > secure file reads > reads from a validated Windows ACL and owner`. Node **24** `windows-latest` passed on the identical commit.

That test is not reachable from this change — the diff touches `src/sidecar-lock.ts` and `src/sidecar-lock-policy.ts`, and neither `readSecureFile`, `inspectPathPermissions`, nor the owner/DACL path takes a file lock. It is also a repeat: the same test failed on #82 and #85.

Worth noting what *did* change: the two failures this PR is about — `file-lock-reentrancy.test.ts` and `new-primitives.test.ts > file locks` — passed on every Windows leg here, including the Node 22 leg that went red for the unrelated ACL timeout.

The remaining ACL failure looks like a different defect (a 15s timeout rather than an `EPERM`), so I have left it out of this PR rather than widening the scope. Happy to chase it separately if useful.

The first run of this branch did have two genuine failures of my own, both in the new tests — they held the raw `mkdtemp()` result while `acquire()` realpaths the target, so on a Windows runner's 8.3 short name the injected path never matched. Fixed in `0dcf0c4`, verified locally against a directory junction that reproduces the same mismatch.

Yigtwxx:
The P1 is correct. `f02f970` narrows it.

The classification sat in the catch wrapping the whole acquisition body, which also covers `options.payload()`, `lockRoot.create/open()`, the payload write and the stat. A Windows `EPERM` from any of those was reclassified as contention — the loop reran the caller's payload callback, and once the retry budget was spent `waitForRetry()` replaced the caller's error with `file_lock_timeout`. For a fail-closed boundary that is the wrong direction, and my evidence never covered those paths.

Classification now happens at the two operations the evidence does cover:

- the direct exclusive create, via a `try` around `createNativeExclusiveFile() ?? fs.open(lockPath, "wx")` that records `lockFileCreateDenied` and rethrows
- the holder snapshot read, which was already in its own `try`

The loop only owns the budget now (`withinDenialBudget()`), so nothing else in the acquisition body can be reclassified. The `lockRoot` branch is deliberately left alone — it routes through `Root` and raises `FsSafeError`, and I have no evidence of the teardown window there.

Added the regression you asked for: an `EPERM` thrown by the caller's `payload` with `retry: { retries: 0 }` now rejects as `EPERM` rather than `file_lock_timeout`, and the callback is asserted to run exactly once. It fails on the previous commit and passes on this one.

Re-verified after the narrowing rather than assuming it held: `test/file-lock-reentrancy.test.ts` + `test/new-primitives.test.ts` + `test/sidecar-lock-regression.test.ts`, **0 failures in 60 runs** on Windows 11 / Node 22. Full local `vitest run` is 490 passed / 183 skipped, `tsc --noEmit` clean, both lint scripts exit 0.

On the P1 merge risk about the hosted Windows check: that failure was `new-primitives.test.ts > reads from a validated Windows ACL and owner` timing out at 15s, which is not reachable from this diff — no lock is taken anywhere in `readSecureFile`, `inspectPathPermissions`, or the owner/DACL path. It is a separate spawn-budget problem and I opened #88 for it with the measurements. The two tests this PR is about passed on every Windows leg here, including the leg that went red for that timeout.

Yigtwxx:
Also correct. `3d0ffe4`.

`createNativeExclusiveFile()` opens `dirname(lockPath)` before its native exclusive open, so a denial from that setup step was still landing inside the classified `try` — the same fail-closed violation as the first finding, one level further in.

The classification now requires the error to **name the lock file**:

```ts
export function isTransientLockFileDenial(error: unknown, lockPath: string): boolean {
  const denial = error as NodeJS.ErrnoException | null;
  return process.platform === "win32" && denial?.code === "EPERM" && denial.path === lockPath;
}
```

That maps onto the evidence rather than onto a code path. Every observed failure carries `path` set to the lock file — the CI stack was `{ errno: -4048, code: 'EPERM', syscall: 'open', path: '...\state.json.lock' }`, from `fs.open(lockPath, "wx")`, and the snapshot read reports the same. A parent-directory denial names the directory and propagates untouched.

On covering the native-enabled case: a native `openBeneath` failure that identifies no path is **not** retried. I would rather leave that propagating than widen the rule past what I can demonstrate — I have no evidence the teardown window is reachable through the native create, and failing closed is the right default on this boundary. If you want it covered, the honest way is to tag the error inside `createNativeExclusiveFile` so the create step is distinguishable from its setup, and I am happy to do that as a follow-up with a native-enabled reproduction behind it.

New regression: a denial naming the lock **parent** rejects with the original `EPERM` and runs the payload callback exactly once. It fails on `f02f970` and passes here. That is alongside the earlier payload-`EPERM` regression, so both escape routes out of the classified block are now pinned.

Re-ran the race after the change rather than assuming the discriminator still matched: `file-lock-reentrancy` + `new-primitives` + `sidecar-lock-regression`, **0 failures in 60 runs** on Windows 11 / Node 22. Full local `vitest run` 491 passed / 183 skipped, `tsc --noEmit` clean, both lint scripts exit 0. The previous head also went fully green on the hosted matrix (21 checks, no failures), including every Windows leg.

Yigtwxx:
Correct again, and this one contradicted a claim in my own PR body. `4603817`.

`withinDenialBudget()` bounded only the denial count. The denial still went through `waitForRetry()`, which throws `file_lock_timeout` under `retries: 0` or an elapsed deadline — so on exactly the settings that ask to fail fast, the caller lost the permission diagnosis. I wrote that bounded denials keep `EPERM` surfacing; they did not.

The wait now hands the original denial back when no further attempt can be scheduled:

```ts
const retryOrRethrowDenial = async (denial: unknown): Promise<void> => {
  try {
    await waitForRetry();
  } catch (waitError) {
    if ((waitError as NodeJS.ErrnoException).code === "file_lock_timeout") throw denial;
    throw waitError;
  }
};
```

Both the create and snapshot-read branches route through it, and any other `waitForRetry()` failure still propagates as itself.

New regression: a lock-file `EPERM` with `retry: { retries: 0 }` rejects as `EPERM` with `path` set to the lock file. It fails on `3d0ffe4` with `code: 'file_lock_timeout'` and passes here. That is the fourth case pinned on this loop, alongside the payload denial, the parent denial, and the two teardown retries.

Re-ran the race again rather than assuming: **0 failures in 60 runs** on Windows 11 / Node 22 across `file-lock-reentrancy`, `new-primitives` and `sidecar-lock-regression`. Full local `vitest run` 492 passed / 183 skipped, `tsc --noEmit` clean, both lint scripts exit 0. `3d0ffe4` was also fully green on the hosted matrix, 21 checks with no failures.

One thing worth your judgement rather than mine: the line-budget entries have crept across these rounds — `src/sidecar-lock.ts` to 540 and now a new entry for `test/sidecar-lock-regression.test.ts` at 540. Each round added a guard the review asked for, but the net effect is two raised budgets in `scripts/check-file-size.mjs`. If you would rather see the acquisition loop or the regression file split than the budgets raised, say which and I will do that instead of carrying the allowlist entries.

Yigtwxx:
@clawsweeper re-review

The open P1 (`Preserve EPERM when no retry is permitted`) was addressed in `4603817`; the last review is against `3d0ffe4`. The hosted matrix is green on the current head — 21 checks, no failures.

steipete:
Landed as b1132f9 via #92 — thank you, this was a good catch and the evidence was excellent.

Your diagnosis was right and your five commits are preserved verbatim in the merged history, with the changelog crediting you. I closed the one remaining gap myself rather than sending it back: `createNativeExclusiveFile()` runs before the `"wx"` fallback whenever the binding is loaded, and the Windows binding mapped `ERROR_ACCESS_DENIED` to `EACCES` without a pathname, so your `EPERM`-plus-exact-path predicate could never match on packaged Windows installs. Since every test here forces native mode off, it stayed invisible.

The Rust mapper now reports `ERROR_ACCESS_DENIED` as `EPERM`, which also fixes a latent inconsistency you had already half-surfaced — the native path and the JavaScript fallback were reporting different codes for the identical condition. The target path is attached only on the `openBeneath()` exclusive-create failure, so the predicate stays exactly as narrow as you wrote it and unpathed denials remain unclassifiable.

I also took you up on the module split instead of raising `LINE_BUDGETS`. Both budget exceptions are now gone rather than added to.

The 8-in-85 versus 0-in-110 measurement was the thing that made this reviewable — a single green run would have proved nothing for a race this shape. Thanks for doing it properly.

PR #88, opened 2026-08-02
test(permissions): budget the Windows ACL tests for the no-native path

Yigtwxx:
## What Problem This Solves

`new-primitives.test.ts > secure file reads > reads from a validated Windows ACL and owner` times out at 15s on Windows CI. It has failed that way on #82, #85 and #87, always on one Node version while the other passes on the identical commit. The affected surface is test reliability for the Windows permissions fallback — nothing ships differently.

The cause is that the `Node N check` job never runs `pnpm native:build`, so these tests run the *command* fallback rather than the native `readOwnerAndDacl()`. Instrumenting `defaultPermissionExec` shows the cost:

```
SPAWNS inspect=2 total=4 ms=546
LOG ["powershell.exe -NoLogo -NoProfile -NonInteractive -Enco",
     "icacls.exe C:\...\secret.json",
     "powershell.exe -NoLogo -NoProfile -NonInteractive -Enco",
     "icacls.exe C:\...\secret.json"]
```

Each `inspectPathPermissions()` spawns `powershell.exe` for the owner query (`inspectWindowsOwner`) plus `icacls.exe` for the ACL. Each test pays **six process spawns**: two `icacls` in `secureWindowsTestFile`, then two inspecting calls at two spawns each.

## Why This Change Was Made

This is a timing budget for a spawn-bound test, not a slow assertion, so the budget is the thing that is wrong. Both Windows ACL tests do identical work and both move from 15s to 60s.

Cold process starts dominate on a Windows runner. The same failing run showed the machine five to ten times slower than local across unrelated suites — `api-coverage` at 6833ms and `archive` at 5306ms — and a test whose cost is almost entirely `CreateProcess` scales worse than wall-clock I/O does. Locally these tests take 561ms.

Non-goals: this does not change the library. `permissions.ts` still prefers `inspectWindowsPermissionsNative()`, which performs no spawns, and only falls back to the commands when no binding is loaded. Jobs that do build the binding never pay this cost.

## User Impact

None for package consumers. Test-only change.

## Evidence

Local run of the affected file, all green:

```
Test Files  1 passed (1)
     Tests  59 passed | 11 skipped (70)
```

Full local verification: `vitest run` 487 passed / 183 skipped, `tsc --noEmit` clean, `scripts/check-file-size.mjs` and `scripts/check-fs-boundary-primitives.mjs` both exit 0.

The spawn counts above come from temporarily instrumenting `defaultPermissionExec` to record each command; that instrumentation is not part of this diff.

### Separate finding, not fixed here

`defaultPermissionExec` in `src/permissions.ts` calls `execFileAsync` with no `timeout`, and so does `secureWindowsTestFile`. A wedged `icacls.exe` or `powershell.exe` would hang `inspectPathPermissions()` indefinitely for a consumer, not just in tests. That is a real robustness gap but it is a behaviour change to public API, so I have left it out of a test-only PR. Happy to open it separately if you want it bounded.

- [ ] `CHANGELOG.md` updated when release-relevant — test-only, nothing ships

### DIFF
--- a/test/new-primitives.test.ts
+++ b/test/new-primitives.test.ts
@@ -239,7 +239,7 @@ describe("secure file reads", () => {
         ownerTrusted: true,
       });
     },
-    15_000,
+    60_000,
   );
 
   it.runIf(process.platform === "win32")(
@@ -268,7 +268,7 @@ describe("secure file reads", () => {
         ownerTrusted: true,
       });
     },
-    15_000,
+    60_000,
   );
 
   it("rejects symlinks and files outside trusted dirs", async () => {


Yigtwxx:
`Node 24 check (ubuntu-latest)` is red on this branch. It is not from this diff — and I do not think it should be written off as noise either.

This PR changes two numbers, `15_000` to `60_000`, on tests guarded by `it.runIf(process.platform === "win32")`. On ubuntu those tests do not run, so the timeout value has no effect there.

What failed:

```
AssertionError: expected [ 'logs', 'secret.txt' ] to deeply equal [ 'secret.txt' ]
 ❯ test/write-boundary-bypass.test.ts:336:51
```

That is `does not create directories outside the root when copyIn fallback races a parent symlink swap`. The assertion found a `logs` directory **outside** the root, which is the exact escape the test exists to catch: the swapper replaces `data` with a symlink to `layout.outside` while `copyIn("data/logs/2026/07/app.log")` runs, and one of the 400 attempts appears to have created the parent chain through the swapped symlink. The test sets `configureFsSafeNative({ mode: "off" })`, so this is the command/JS fallback rather than the native path.

Caveats on how far I can take that claim:

- One observation. I checked recent `main` runs and the ubuntu failure there was `json.test.ts:329` (`expected 9 to be greater than 10`), a different test, so this is not obviously the same known-flaky family.
- I could not reproduce it. The boundary-bypass tests are `runIf(process.platform !== "win32")` and I am on Windows, so I have no way to run the racing case locally or to tell an intermittent real escape from a harness artifact.

Flagging rather than diagnosing, since a genuine intermittent containment escape in the non-native fallback would matter well beyond this PR. Anyone with Linux CI can hit it far more cheaply than I can from here.

steipete:
Superseded by #89, merged as 64f7742 — thank you, your diagnosis was the whole basis for it.

I didn't take the timeout increase, but only because your own analysis pointed at something better. You correctly identified that the `Node N check` job never runs `pnpm native:build`, so those tests exercise the command fallback and pay six process spawns each. Rather than widen the budget to fit that cost, #89 removes a redundant inspection so the affected tests drop to four spawns, and keeps ordinary Windows CI on the real command fallback — that path is real shipping code and deserves the coverage it was getting by accident.

The more important part is the finding you flagged and deliberately left out of a test-only PR: `defaultPermissionExec` calling `execFileAsync` with no `timeout`. You were right that it is a real robustness gap, and right that it was bigger than the flake. A wedged `icacls.exe` or `powershell.exe` would hang `inspectPathPermissions()` indefinitely for any consumer. It is now bounded to 10 seconds and fails closed — an owner query that cannot complete yields `source: "unknown"`, the ACL query is skipped, and `readSecureFile()` refuses with `permission-unverified` rather than returning a permissive answer.

The affected test went from 15-second timeouts to about four seconds on the merged branch. Thanks for chasing the cause instead of just the symptom, and for scoping the PR honestly rather than smuggling the API change in with it.

PR #109, opened 2026-08-03
fix: guard Root fallback cleanup by identity

steipete:
## Summary

- exercise Root mutation failures, pinned-path decisions, JavaScript fallbacks, deny-mutation policy, absolute-path classification, canonical root paths, and private/sync FileStore failures
- preserve unrelated files when create-only fallback verification loses a parent-directory race by cleaning up only the inode created by the operation
- normalize writable directory failures to the documented not-file code across Windows and POSIX
- assert stable FsSafeError codes and boundary state after every injected failure

## Bugs fixed

The create-only JavaScript Root fallback previously cleaned up a failed write with a pathname-only removal. A deterministic parent swap during post-write verification redirected that cleanup through a symlinked parent and deleted an unrelated attacker file outside the root. The operation already rejected with path-mismatch; cleanup crossed the boundary afterward.

The fallback now records the created inode before writing and reuses the identity-checked guarded cleanup path. The regression proves the outside file survives while the library-created file is preserved under the displaced original parent.

The new cross-platform mutation matrix also exposed that Windows could reach the writable descriptor stat check for a directory and report invalid-path, while POSIX rejected the same target through EISDIR as not-file. The shared descriptor check now reports not-file, matching the public error contract on both platforms.

## Coverage

Native-backed local measurement, statements / branches / functions / lines:

| File | Before | After |
| --- | --- | --- |
| absolute-path.ts | 77.24 / 63.73 / 87.50 / 77.57 | 91.01 / 80.21 / 100 / 90.90 |
| root-paths.ts | 83.33 / 71.29 / 91.66 / 83.33 | 92.47 / 86.11 / 100 / 92.47 |
| file-store.ts | 87.73 / 77.08 / 95.23 / 87.73 | 96.93 / 88.88 / 97.61 / 96.93 |
| root-impl.ts | 86.85 / 79.64 / 74.75 / 88.23 | 92.04 / 85.92 / 79.61 / 93.11 |
| root-path.ts | 93.84 / 85.71 / 100 / 93.78 | 93.84 / 85.71 / 100 / 93.78 |
| root-context.ts | 95.45 / 86.48 / 85.71 / 95.45 | 100 / 89.18 / 100 / 100 |
| whole project | 86.03 / 78.01 / 87.24 / 87.71 | 87.21 / 79.40 / 88.15 / 88.90 |

## Defensive branches not contrived

- The absolute ancestor walker cannot reach its null/root-exhaustion return on a supported mounted filesystem because the filesystem root exists; its downstream no-ancestor canonical fallback is therefore defensive.
- The absolute directory prefix loop cannot observe an empty segment after filtering empty segments.
- Pinned write resolution cannot produce an absolute relative path after the upstream root resolver has already enforced containment, and its later opened-path/outside-root checks duplicate earlier open confinement.
- The move fallback's source-symlink branch is preempted by strict pinned source resolution. The rename EEXIST mapping is Windows-specific and is not naturally produced by rename on the Linux coverage host.

## Verification

- pnpm check
- pnpm test:security
- pnpm native:build && pnpm test:coverage
- git diff --check
- Codex autoreview: clean twice, no accepted/actionable findings


steipete:
Landed as `1d8f4224c92ab4b22f6b1577e6cdc05da6670d70` after integration against current `main`.

Verification:

- `pnpm check`: 93 files, 1,013 tests; 954 passed and 59 platform/optional cases skipped.
- `cargo test --workspace --locked`: 16/16 passed.
- Built-library parent-swap probe: `errorCode=path-mismatch`, outside replacement remained `attacker`, and the displaced original contained `library`.
- Final autoreview: TruffleHog clean; no accepted/actionable findings.
- Exact-head GitHub runs at `28e76e707563879e5612ebc802dbccfd67299412`: CI, coverage, CodeQL, benchmarks, and ClawSweeper dispatch all succeeded.

The first integrated CI run exposed a separate Windows stress-test timeout from #113. The focused test-only timeout adjustment was autoreviewed and the full exact-head matrix then passed, including Windows Node 22 and Node 24.

PR #122, opened 2026-08-05
fix(ci): stabilize Windows race regressions

steipete:
## Summary

- inject the retryable `path-mismatch` failure directly when testing JSON retry exhaustion, preserving the exact five-attempt budget without depending on Windows rename identity timing
- retry only transient Windows directory-rename denials while arranging the real parent-swap regression
- propagate post-`lstat` test-hook failures so a failed race setup cannot be silently treated as an ordinary path inspection error

## Proof

- `pnpm test test/json.test.ts test/path-stress-regression.test.ts` (10 repeated runs)
- `pnpm check` (100 files passed, 1,035 tests passed, 61 skipped)
- `git diff --check`
- Codex autoreview: clean, no accepted/actionable findings

Fixes the Windows failures in coverage run 30976077742.

### DIFF
--- a/src/root-impl.ts
+++ b/src/root-impl.ts
@@ -219,12 +219,14 @@ async function openVerifiedLocalFile(
     if (!preOpenStat.isFile() && !preOpenStat.isSymbolicLink()) {
       throw new FsSafeError("not-file", "not a file");
     }
-    await fsSafeTestHooks?.afterPreOpenLstat?.(filePath);
   } catch (err) {
     if (err instanceof FsSafeError) {
       throw err;
     }
     // ENOENT and other lstat errors: fall through and let fs.open handle.
+  }
+  if (preOpenStat) {
+    await fsSafeTestHooks?.afterPreOpenLstat?.(filePath);
   }
 
   let handle: FileHandle;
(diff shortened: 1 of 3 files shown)


steipete:
Landed as `8cc745a20b1d19340b27aaf1e73b1247a02297cd`.

Verification:

- `pnpm test test/json.test.ts test/path-stress-regression.test.ts` passed across 10 repeated local runs.
- `pnpm check` passed: 100 test files, 1,035 tests passed, 61 skipped, plus build, docs, filesystem-boundary lint, file-size lint, and package checks.
- `git diff --check` passed.
- Codex autoreview was clean before commit and again immediately before landing.
- Exact-head CI at `1307bf321daee2d52013c602dff5c924edf317da` passed the full CI, CodeQL, benchmarks, and coverage matrices, including Node 22 coverage on Windows, Windows Node 22/24 checks, native Windows checks, and Windows package smoke.

One unrelated Ubuntu coverage attempt timed out in `native-publish-equivalence.test.ts`; rerunning the failed job passed, and merged coverage then passed. No verification images applied to this test-only CI fix.

PR #127, opened 2026-08-10
fix(path): reject raced FIFOs without blocking

obviyus:
## What Problem This Solves

Fixes an issue where pinned and standalone regular-file reads could block indefinitely when a regular path was replaced by a FIFO or device between inspection and descriptor open.

## Why This Change Was Made

All affected readers now use one internal POSIX nonblocking, no-follow flag resolver before descriptor type and identity validation. Root reads, hashing, and move-copy fallbacks share the same policy. No public interface or error shape changes.

## User Impact

Raced special files are rejected promptly instead of pinning a worker. Ordinary regular-file reads retain their existing behavior.

## Evidence

- Before: the new pinned, async, and sync FIFO-swap regressions failed because descriptor flags contained `0` instead of `O_NONBLOCK` (`2048`).
- After: `pnpm check` passes: build, docs, 1,035 tests, package validation, and filesystem-boundary lint.
- Autoreview: clean; no accepted or actionable findings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Security and Correctness
 
+- Open pinned and standalone regular-file reads nonblocking on POSIX before descriptor type validation, so a raced FIFO or device cannot stall the worker; share the same read-open flags with root reads, hashing, and move-copy fallbacks.
 - Compare exact bigint filesystem identities during exclusive publication and rollback cleanup, so rounded Windows file indexes cannot hide replaced source or target paths or authorize deletion of an attacker replacement.
 
 ## 0.5.3 - 2026-08-08
--- a/src/file-hash.ts
+++ b/src/file-hash.ts
@@ -1,10 +1,10 @@
 import { createHash } from "node:crypto";
-import fsSync from "node:fs";
 import type { FileHandle } from "node:fs/promises";
 import fs from "node:fs/promises";
 import { FsSafeError } from "./errors.js";
 import { sameFileIdentity } from "./file-identity.js";
 import { getNativeBinding, type NativeBinding } from "./native.js";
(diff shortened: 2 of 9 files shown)


obviyus:
Exact-head POSIX FIFO swap proof (`c8b98e0ac9a0ce994af726719e6a6359ab7f3ed8`):

The test renames the regular file after `lstat`, creates a real FIFO with `mkfifo`, then calls the real `fs.openSync` with the production flags. The descriptor is rejected as non-regular without waiting for a writer.

```text
$ pnpm test test/pinned-open.test.ts -t 'rejects a raced FIFO without blocking the descriptor open' --reporter=verbose
RUN v4.1.10 <checkout>
✓ openPinnedFileSync > rejects a raced FIFO without blocking the descriptor open 4ms
Test Files  1 passed (1)
Tests       1 passed | 7 skipped (8)
Duration    127ms (tests 5ms)
elapsed=0.49 sec exit=0
```

obviyus:
Maintainer decision: approve merge. The shared read-open flag resolver is the correct owner boundary; it preserves descriptor validation while removing the blocking FIFO race across every guarded reader.

obviyus:
Landed via squash onto `main`.

- Exact-head ClawSweeper: PASS; 0 findings, 0 security concerns, no maintainer decision, 0 Rank-up moves
- Validation: `pnpm check` plus green Linux, macOS, and Windows CI
- Runtime proof: real POSIX FIFO swap rejected in 4 ms
- Changelog: `CHANGELOG.md` updated
- Land commit: `c8b98e0ac9a0ce994af726719e6a6359ab7f3ed8`
- Merge commit: `99f917e55be4902094e6fbd53a2de27cce120fcc`

PR #140, opened 2026-08-23
fix(lock): apply configured lock defaults to synchronous acquires

Yigtwxx:
Synchronous file locks silently ignored `configureFsSafeLocks()` even though it defines process-wide sidecar defaults. This PR resolves `retry`, `staleMs`, `staleRecovery`, and `timeoutMs` at synchronous acquisition, matching the async wrapper: per-call values win, configured zero values survive, and per-call retry objects replace the configured object as a whole. `withFileLockSync()` delegates to that owner.

Stale recovery still requires effective `remove-if-unchanged`, caller approval through `shouldRemoveStaleLock`, an exclusive reclaim guard, and unchanged-snapshot deletion. A per-call `fail-closed` override prevents recovery even with global opt-in. Applying existing process configuration to synchronous callers is an intentional compatibility-visible correction; it does not grant recovery authority by itself.

Synchronized head: `b32306320954daca4266b64c07eac87aef9cd658`, merging main `f0665cf0270f62d54e23b0cabfe7b09d902b92ff` after #141 and #142. Only the changelog conflicted; all credited entries were preserved. The five-file net PR patch is identical to approved head `dfe3681d7f831e88923fa28e1ad140475980636e` against its original base, excluding line offsets/context. Lock source, configuration tests, and both documentation pages are byte-identical to that approved head. All other files match new main, retaining landed archive, identity, staging, write, and native-smoke fixes. No unlanded documentation changes are stacked. Original contributor history and credit are preserved.

The 27 configuration cases cover configured timeout under sidecar, in-process-holder, reclaim-guard, and wrapper contention; retry/backoff/stale policy; per-call overrides and zero values; whole-object retry replacement; missing/false approval, fail-closed overrides, and replacement preservation. Real temporary lock files and a controlled clock/sleep with a hard 12-sleep escape ensure a synchronous regression fails rather than blocking the test worker. Configuration, mocks, and fixtures are restored.

Existing timing differences are unchanged: synchronous retry sleeps are not clamped to the remaining deadline, and infinite timeout retains the synchronous retry cap. The original timestamped retry fixture would not necessarily wait forever without its explicit timeout because its timestamp can become stale. Prior controlled regression proof at base `ffd07ff2d0d2dcf5482574de963016f1c513f70e` produced 17 failures / 10 passes, then 27 passes after restoring the approved implementation. That historical experiment was not repeated during this synchronization.

Fresh local validation on Node 26.7.0 / pnpm 11.24.0:

- Focused sync configuration/failure/exit, sidecar failure/reclaim/ownership, reentrancy, and concurrency suites: **113 passed, 8 platform skips**. The initially mistyped reentrancy selector was corrected with a separate `pnpm test test/file-lock-reentrancy.test.ts` run; its 7 passes are included in the total.
- `pnpm test:security`: **78 passed**.
- `pnpm check`: lint, build, and documentation checks passed; full tests reported **1,601 passed, 224 skipped, 2 timeout failures**. The unchanged archive-fuzz malformed-header case exceeded 15 seconds; the unchanged non-private store-stress reader/writer case exceeded 5 seconds. Host load was elevated, but the normal full check is still recorded as failed.
- The single bounded retry, `pnpm test test/archive-property-fuzz.test.ts test/store-stress.test.ts --maxWorkers=1`, passed **25 tests, 2 skips**, without changing test limits or assertions. No second full-suite attempt was made.
- Separate `pnpm pack:check`, `git diff --check`, and the base-relative whitespace check passed.
- One official isolated Codex autoreview with `--mode local --base origin/main` and high effort reviewed the complete synchronized candidate: no findings at the skill's default **P0-only** scope. Credential scanning and the restricted reviewer were retained.

Fresh hosted verification passed for the synchronized head, with no hosted failures, reruns, or CI repairs:

- [CI](https://github.com/openclaw/fs-safe/actions/runs/33217928295): all 15 jobs passed, including Node 22/24 full checks on Linux/macOS/Windows, native checks on those platforms and musl, bundled-package smoke, workflow lint, and Cargo quality.
- [Coverage](https://github.com/openclaw/fs-safe/actions/runs/33217928285): all three operating systems and merged thresholds passed.
- [CodeQL](https://github.com/openclaw/fs-safe/actions/runs/33217928289), [benchmarks](https://github.com/openclaw/fs-safe/actions/runs/33217928268), and [dispatch](https://github.com/openclaw/fs-safe/actions/runs/33217926287) passed.

Canonical Actions run and job metadata confirms the exact head and reviewed tree `9d55546c4b6d6676001a0f7bcc8a19f0a97eeb77`. Native OS jobs successfully built the package, staged their binding, passed required-mode smoke, and ran native archive/security and Root publication checks; the musl build/test step also passed. The only skipped check at verification was the optional `[code]smith` automation.

No local native-required proof is claimed. Native/archive code is inherited unchanged; the hosted native lanes above supply fresh proof. PR140 remains open with auto-merge off, awaiting maintainer approval of this changed head. No merge or publication has been performed.


Yigtwxx:
The open item is the compatibility decision, so here is the blast radius measured rather than argued.

**`configureFsSafeLocks()` has no caller anywhere in the org.** GitHub code search across `openclaw` returns hits only inside fs-safe itself — `src/lock-config.ts`, `src/index.ts`, `src/config.ts`, `src/advanced.ts`, `docs/config.md`, `docs/json-store.md`, `test/sidecar-lock-regression.test.ts`, `test/public-api.json`. No product code sets a process-wide lock default today, which means `defaults.staleMs`, `defaults.retry` and `defaults.timeoutMs` are all `undefined` in practice and the only field that resolves to anything is `staleRecovery`, whose package default is `"fail-closed"` — the value the synchronous path already assumed.

**The one real consumer of the synchronous API passes its options per call.** `acquireFileLockSync` is used by `openclaw/openclaw` through a single wrapper, `src/infra/file-lock-sync.ts:22-29`:

```ts
const lock = acquireFileLockSync(path, {
  staleMs: 30_000,
  retry: { retries: 9, factor: 1, minTimeout: 20, maxTimeout: 20, randomize: false },
  staleRecovery: "remove-if-unchanged",
  payload: createPayload,
  shouldReclaim: isStale,
  shouldRemoveStaleLock: isStale,
});
```

Every field this PR touches except `timeoutMs` is supplied explicitly, and per-call options win before and after the change. `timeoutMs` is not passed, so it would newly resolve from process configuration — except there is none to resolve from, per the search above. Both call sites that reach this wrapper (`src/agents/sessions/settings-storage.ts`, `src/commands/doctor-auth-migration-receipts.ts`) go through it, so the observable behavior in the largest downstream consumer is unchanged for two independent reasons.

That also means the `remove-if-unchanged` half of the risk cannot fire from configuration alone in any code that exists today: it needs a process default *and* a per-call `shouldRemoveStaleLock`, and the only code supplying the second one already supplies the first per call.

I would still rather you accept it explicitly than have me narrow it, because narrowing is the wrong direction here. Dropping `staleRecovery` from the fill would leave `configureFsSafeLocks()` half-honored on the synchronous path — three fields applied, one silently ignored — which is a worse contract than either the old behavior or this one. If you would prefer the conservative shape anyway, the smallest version is to fill `staleMs`, `retry` and `timeoutMs` only and document `staleRecovery` as async-only in `docs/sidecar-lock.md:172`; say the word and I will push that instead.

Method note so the measurement is checkable: `gh search code "configureFsSafeLocks" --owner openclaw` and the same for `acquireFileLockSync` / `withFileLockSync`. Code search indexes default branches only, so a caller on an unmerged branch would not appear.

steipete:
Landed as 209ad78294f0338c1b37cc892c6fd34c39459e7e from approved head b32306320954daca4266b64c07eac87aef9cd658. The intentional compatibility correction is accepted: synchronous locks now honor all four configured defaults, preserving per-call precedence, explicit zeros, and whole-object retry replacement. Configured stale recovery still requires per-call approval, an exclusive reclaim guard, and an unchanged snapshot; per-call fail-closed disables it. Thanks @Yigtwxx; the squash commit includes explicit co-author credit.

Post-merge verification confirmed the merge is reachable from origin/main and its complete tree equals the approved candidate. The actual merged source, tests, docs, and changelog were inspected: five files, +253/-4. All files outside this scope match prior main, preserving the landed ZIP/TAR fixes and native smoke handling. `git diff --check` and the merged-diff whitespace check passed.

Recorded proof for this exact source head (macOS arm64, Node 26.7.0, pnpm 11.24.0; reused during landing, not rerun):

```sh
pnpm test test/file-lock-sync-config.test.ts test/file-lock-sync-failure.test.ts test/file-lock-sync-exit.test.ts test/lock-reentrancy.test.ts test/sidecar-lock-acquire-failure.test.ts test/sidecar-lock-regression.test.ts test/sidecar-lock-helpers-failure.test.ts test/sidecar-lock-ownership-token.test.ts test/sidecar-lock-windows-denial.test.ts test/file-lock-concurrency-stress.test.ts test/json-store-concurrency.test.ts
pnpm test test/file-lock-reentrancy.test.ts
pnpm test:security
pnpm check
pnpm test test/archive-property-fuzz.test.ts test/store-stress.test.ts --maxWorkers=1
pnpm pack:check
```

The first focused command passed 106 tests with eight platform skips; its nonexistent `test/lock-reentrancy.test.ts` selector was ignored. The corrected second command passed seven reentrancy tests, for 113 focused passes total. Security passed all 78 tests. Lint, build, and docs checks passed, but the full local check was **not green**: 1,601 passed, 224 skipped, and two timeouts (archive fuzz at 15 seconds; non-private store stress at five seconds). Exactly one bounded retry of those unchanged files passed 25 tests with two skips; no test limits or assertions were relaxed. The separate package check passed. No local native-required proof is claimed for this lock change.

At the final merge gate, checks were 25 SUCCESS and three benign SKIPPED: optional [code]smith plus two bot-metadata dispatch jobs. Exact-head [CI](https://github.com/openclaw/fs-safe/actions/runs/33217928295), [coverage](https://github.com/openclaw/fs-safe/actions/runs/33217928285), [CodeQL](https://github.com/openclaw/fs-safe/actions/runs/33217928289), and [benchmarks](https://github.com/openclaw/fs-safe/actions/runs/33217928268) passed without reruns. Hosted CI includes Node 22/24 checks across Linux/macOS/Windows, native OS and musl lanes, and bundled-package smoke; those are hosted proofs, distinct from local results. The clean isolated review was P0-only. One write preflight stopped safely while the bot updated its review; the update reported no code/security findings and repeated the compatibility decision accepted above. No post-merge CI watcher was started.

PR #164, opened 2026-08-30
fix(archive): join timeout cleanup before rejection

steipete:
## Summary

- preserve the released prompt-timeout behavior for non-mutating archive work
- track ownership only while a live destination mutation and its rollback are in flight
- prevent any new destination mutation from starting after the deadline across JavaScript ZIP/TAR and native extraction
- add deterministic JavaScript/native ZIP/TAR regressions for both prompt rejection and in-flight mutation joining

## Root cause

The public timeout used `Promise.race()`, while the staged-to-destination merge did not observe the deadline between awaited guards and live mutations. A merge paused before destination creation could therefore outlive the public rejection, resume later, and overwrite trusted recovery content.

The fix keeps the hard timeout path for non-mutating work. Destination mutations now register with the deadline owner only for their live mutation/validation/rollback scope. On timeout, the public promise waits only when such a scope is already active; otherwise it rejects promptly. Checks after every awaited pre-mutation boundary prevent an abandoned merge from starting publication later.

## Live behavior proof

A fresh npm consumer installed the packed current package and invoked the public `extractArchive()` API on Node 24.20.0/npm 11.19.0. The probe paused the live destination validation boundary through Node's public filesystem module, waited for a 500 ms timeout, wrote trusted recovery content after rejection, released the abandoned merge, and verified both the recovery bytes and private staging cleanup.

JavaScript fallback:

```json
{"mode":"off","timeoutMessage":"extract zip timed out after 500ms","elapsedMs":502,"rejectedBeforeRelease":true,"content":"trusted-recovery","stagingClean":true,"stagedDirCount":2}
```

Native-required path:

```json
{"mode":"require","timeoutMessage":"extract zip timed out after 500ms","elapsedMs":512,"rejectedBeforeRelease":true,"content":"trusted-recovery","stagingClean":true,"stagedDirCount":2}
```

## Verification

- `CI=1 pnpm check` — 133 test files passed, 1 skipped; 2,194 tests passed, 25 skipped
- all archive-prefixed tests — 15 files, 453 tests passed
- focused timeout lifetime suite covers prompt non-mutating rejection, in-flight mutation joining, and JavaScript/native ZIP/TAR post-timeout publication rejection
- packed npm consumer proof above passed in native `off` and `require` modes
- Codex autoreview — clean, no accepted/actionable findings
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Security and Correctness
 
-- Keep archive timeout failures joined to extraction ownership, stopping new destination mutations at the deadline and waiting for in-flight publication and staging cleanup before rejecting.
+- Stop archive destination publication at the timeout boundary and join only an in-flight destination mutation and rollback before rejecting, preserving prompt deadlines for non-mutating work.
 - Verify durable JSON queue reads with lossless bigint identities and reject persistent unknown Windows identities before reading entry bytes.
 
 ## 0.6.0 - 2026-08-29
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Security and Correctness
 
+- Keep archive timeout failures joined to extraction ownership, stopping new destination mutations at the deadline and waiting for in-flight publication and staging cleanup before rejecting.
 - Verify durable JSON queue reads with lossless bigint identities and reject persistent unknown Windows identities before reading entry bytes.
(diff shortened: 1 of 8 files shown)


steipete:
@clawsweeper automerge

steipete:
@clawsweeper stop

steipete:
@clawsweeper automerge

PR #165, opened 2026-08-30
fix(lock): surface asynchronous release failures

steipete:
## Summary

- propagate asynchronous deletion failures when releasing an unchanged, owned sidecar lock instead of reporting success while leaving it behind
- retain exact failed-release state so the same handle, `heldEntries()` force release, or manager drain can retry cleanup safely
- preserve reentrant reference counts across a failed final release and wait for cleanup already in flight before admitting a new reentrant handle
- preserve paired callback/acquisition and cleanup failures with a portable `SuppressedError` shape on Node 22 and newer runtimes

## Root cause

`removeSidecarLockIfUnchanged()` caught and discarded both raw `fs.rm()` and `Root.remove()` failures, then returned `true`. The manager had already removed the lock from its in-process state, so `release()` resolved even though the sidecar remained and could block every later acquirer.

Removal failures now propagate while the handle and manager retain the exact identity receipt. A retry does not decrement reentrant references twice; if another same-owner acquisition arrives after a failed final cleanup, that acquisition keeps the lock live and the old handle's retry cannot remove it. New reentrant acquisition waits for cleanup already in flight, then either creates a fresh lock after success or reuses the retained owner state after failure.

When protected work or acquisition has already failed, a cleanup failure becomes the primary `error` and the original failure remains available as `suppressed`. Node versions with the global `SuppressedError` constructor use it directly; Node 22 receives the equivalent named `Error` shape without raising the package engine floor.

## Live behavior proof

A fresh npm consumer installed the packed current package and exercised the public lock APIs with fail-once deletion errors:

```json
{"direct":{"name":"Error","message":"direct release deletion failed","code":"EIO","lockExistsAfterFailure":true,"lockExistsAfterRetry":false},"combined":{"name":"SuppressedError","error":"combined release deletion failed","suppressed":"protected callback failed","lockExistsAfterFailure":true,"lockExistsAfterDrain":false}}
```

The direct handle rejected with the real deletion error, retained its lock, then removed it on the same handle's retry. `withFileLock()` preserved both errors and retained manager state; `manager.drain()` completed the cleanup after the injected failure was removed.

## Verification

- focused final release/reentrancy suite — 4 files, 40 tests passed
- all lock-related tests before the final reentrancy refinement — 10 files passed, 1 platform suite skipped; 104 tests passed, 8 skipped
- packed npm consumer retry/drain proof above passed
- isolated store stress retry — 8/8 passed after a load-only full-suite timeout
- isolated archive property retry — 34/34 passed after a load-only full-suite timeout
- Codex autoreview — clean on the final combined local/ClawSweeper implementation, no accepted/actionable findings
- `git diff --check`

## CI corrections

The first pushed head passed Node 24 but exposed that Node 22 does not provide the global `SuppressedError` constructor. A portable internal constructor and direct missing-global regression fixed all Node 22 checks. A later Node 24 Windows run had one unrelated archive property-fuzz timeout; its targeted unchanged-code retry passed. The final exact head adds the reviewed retry-state and reentrancy correction and must pass the fresh hosted matrix before landing.


steipete:
@clawsweeper automerge

steipete:
@clawsweeper automerge

steipete:
@clawsweeper automerge

PR #166, opened 2026-08-30
fix(lock): enforce synchronous deadlines

steipete:
## Summary

- clamp synchronous lock backoff to the remaining finite deadline instead of sleeping the full configured delay
- reject invalid async/sync retry counts, factors, delay bounds, and deadline values before filesystem acquisition starts
- preserve explicit zero values and positive infinity as the documented unbounded deadline
- add shared validation plus focused async/sync regressions

## Root cause

The synchronous contention loop checked `timeoutMs`, then slept the complete computed backoff without considering how little deadline remained. A 20 ms deadline with a 200 ms retry therefore returned roughly 200 ms late; an infinite retry delay could block the calling thread indefinitely. Invalid `NaN`, negative, fractional, and infinite retry values also reached `Atomics.wait()` or timer computation directly.

The synchronous path now uses the same remaining-budget clamp as the asynchronous manager. Shared validation runs before either acquisition path mutates the filesystem. Retry counts must be non-negative safe integers, factors and delays must be finite/non-negative, explicit delay ranges cannot be inverted, and lock deadlines must be finite/non-negative or positive infinity.

## Live behavior proof

A real built-package probe contended on an existing sidecar with `timeoutMs: 20` and a 200 ms backoff, then exercised invalid infinite retry delays through both public APIs:

```json
{"timeoutCode":"file_lock_timeout","elapsedMs":21,"syncInvalid":{"name":"RangeError","message":"lock retry.minTimeout must be a finite non-negative number"},"asyncInvalid":{"name":"RangeError","message":"lock retry.minTimeout must be a finite non-negative number"}}
```

The synchronous call returned at the deadline rather than after the 200 ms delay, and both APIs rejected the invalid delay immediately.

## Verification

- all lock-related tests — 11 files passed, 1 platform suite skipped; 123 tests passed, 8 skipped
- focused deadline/validation suites — 4 files; 67 tests passed, 1 skipped
- built-package live proof above passed
- Codex autoreview — clean, no accepted/actionable findings
- typecheck, file-size guard, filesystem-boundary guard, docs examples, and `git diff --check` passed

## Local full-suite note

The full local `pnpm check` reached 2,213 passing tests but hit four unrelated timeout-only failures while many other repository test processes were active on the host: archive timeout lifetime, two store stress cases, and archive property fuzz. The changed lock suites remained green; the archive timeout suite passed 6/6 immediately in isolation, and the complete lock suite passed. Exact-head hosted Linux/macOS/Windows CI remains the landing gate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Security and Correctness
 
+- Validate async/sync lock retry and deadline numbers and clamp synchronous backoff to the remaining finite deadline instead of overshooting or blocking forever.
 - Propagate asynchronous sidecar-lock release and failed-acquire deletion failures, preserve paired errors, and retain failed release state for safe retry through the same handle or manager drain.
 - Stop archive destination publication at the timeout boundary and join only an in-flight destination mutation and rollback before rejecting, preserving prompt deadlines for non-mutating work.
 - Verify durable JSON queue reads with lossless bigint identities and reject persistent unknown Windows identities before reading entry bytes.
--- a/docs/sidecar-lock.md
+++ b/docs/sidecar-lock.md
@@ -94,6 +94,7 @@ type FileLockRetryOptions = {
 ```
 
 `payload` is a function so you can re-evaluate it on each retry (e.g. timestamp, PID).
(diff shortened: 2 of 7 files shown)


steipete:
@clawsweeper automerge

steipete:
@clawsweeper automerge

PR #167, opened 2026-08-30
fix(store): preserve durable queue generations

steipete:
## Summary

- claim each durable-queue generation to an explicit `.processing` path before reading it
- serialize consumers for one ID with a fail-closed cross-process lock and create the claim with an atomic no-replace hardlink
- retire the source through a small crash-recoverable sibling record so producer replacements survive without overwriting the active claim
- make acknowledgement and quarantine operate only on `.processing`, leaving a newer same-ID `.json` replacement untouched and preserving prior failed evidence

## Root cause

The queue reader returned only parsed data while `ackJsonDurableQueueEntry()` and `moveJsonDurableQueueEntryToFailed()` later acted on whichever file currently occupied `<id>.json`. Because queue writes atomically replace that pathname, loading generation 1, publishing generation 2, and then acknowledging generation 1 deleted generation 2.

Loading now serializes consumers for the logical ID through a sidecar lock, creates `<id>.processing` with a no-replace hardlink, and reads only that claimed path. Whichever complete generation occupies `.json` at the hardlink syscall becomes the claim; a producer replacement published afterward remains pending.

Source retirement uses `.fs-safe-retirements/<source-name>/entry`. The current `.json` pathname is atomically moved there and compared with `.processing`: the claimed hardlink is removed, while a producer replacement is restored without replacing a still-newer generation. Empty and populated retirement records are recoverable after a crash. A same-identity `.json`/`.processing` pair left by a crash immediately after hardlink creation is collapsed before reading.

The per-ID lock is intentionally fail-closed and never auto-reclaimed, so a suspended owner cannot resume after another process has advanced the queue. After a hard process crash, an operator must remove the lock only after proving its owner is gone.

Acknowledgement atomically moves only `.processing` through the short-lived `.delivered` marker. Batch loading cleans completed markers before claiming the next pending generation. Acknowledgement now explicitly rejects when `.json` exists without a processing claim: callers that previously paired direct `readJsonDurableQueueEntry()` with acknowledgement must migrate to `loadJsonDurableQueueEntry()` so the acknowledged generation is unambiguous. Missing-state acknowledgement remains idempotent.

Quarantine publishes `.processing` to the failed destination by create-only hardlink, completes an interrupted same-identity hardlink, and rejects an existing different destination while preserving both files. Queue and failed directories therefore need same-filesystem hardlink support, matching the existing same-filesystem rename limitation while adding no-clobber behavior.

## Live behavior proof

A fresh npm consumer installed the packed package and exercised the public store API while replacing the pending generation at the exact hardlink-claim boundary:

```json
{"loaded":{"generation":2},"pending":{"generation":3},"afterAck":{"generation":3},"failed":{"generation":3},"afterQuarantine":{"generation":4}}
```

Generation 2 won the atomic claim, generation 3 remained pending, acknowledgement removed only generation 2, and quarantine moved only the claimed generation 3 while generation 4 remained queued.

## Verification

- focused queue generation/failure/lifecycle suite — 3 files, 32 tests passed
- broad queue/store suite — 10 files, 125 tests passed
- filesystem security suite — 5 files, 78 tests passed
- packed npm consumer proof above passed against the final package
- file-size, filesystem-boundary, typecheck, documentation, and diff checks passed
- Codex autoreview at P1 — clean after the final unclaimed-acknowledgement migration guard
- complete final `CI=1 pnpm check` — 136 files passed, 1 skipped; 2,233 tests passed, 25 skipped; package check passed
- an earlier local attempt had eight unrelated archive/native/store stress timeouts under host load; the clean retry above and all focused reruns passed
- fresh exact-head hosted Linux/macOS/Windows, native, coverage, CodeQL, benchmark, package-smoke, and ClawSweeper checks remain the landing gate


steipete:
@clawsweeper automerge

steipete:
@clawsweeper automerge

steipete:
@clawsweeper automerge

The exact-head repair replaces the check-then-unlink path with a fail-closed per-ID lock, an atomic no-replace hardlink claim, and crash-recoverable source retirement. The PR body now includes the packed-consumer replacement-boundary proof and final local validation.

steipete:
@clawsweeper automerge

The direct-reader compatibility finding is resolved through an explicit safe migration: acknowledgement rejects when pending data exists without a processing claim, preserves that pending generation, and directs callers to load through the generation-bound API first. Missing-state acknowledgement remains idempotent. Focused tests and P1 autoreview are clean.

PR #168, opened 2026-08-30
fix(archive): reject invalid filter policy

steipete:
## Summary

- validate runtime `onFiltered` values through one shared archive-policy resolver
- reject values other than `reject-archive` and `skip-entry` before kind resolution, native selection, staging, or destination mutation
- pass the normalized policy consistently through ZIP, TAR, and native extraction paths

## Root cause

`shouldExtractArchiveEntry()` rejected filtered entries only when `(onFiltered ?? "reject-archive") === "reject-archive"`. Any other runtime value—including an empty string or unknown policy supplied from JavaScript, JSON, or an untyped boundary—fell through to `false`, silently granting skip-entry behavior that is documented as explicit opt-in.

`resolveArchiveFilteredEntryPolicy()` now owns the runtime contract. `extractArchive()` resolves it before any extraction work and forwards only the normalized enum. The shared entry-policy helper also validates directly, so internal or independently tested callers cannot retain the old fallback.

## Live behavior proof

A fresh npm consumer installed the packed package and exercised public ZIP extraction with empty and unknown policy values in native `off` and `require` modes:

```json
[{"mode":"off","onFiltered":"","error":{"name":"RangeError","message":"archive onFiltered must be \"reject-archive\" or \"skip-entry\""},"entries":[]},{"mode":"off","onFiltered":"unknown-policy","error":{"name":"RangeError","message":"archive onFiltered must be \"reject-archive\" or \"skip-entry\""},"entries":[]},{"mode":"require","onFiltered":"","error":{"name":"RangeError","message":"archive onFiltered must be \"reject-archive\" or \"skip-entry\""},"entries":[]},{"mode":"require","onFiltered":"unknown-policy","error":{"name":"RangeError","message":"archive onFiltered must be \"reject-archive\" or \"skip-entry\""},"entries":[]}]
```

Every case rejected before writing destination entries.

## Verification

- focused archive policy/native-equivalence suite — 3 files, 74 tests passed
- complete `CI=1 pnpm check` — 136 files passed, 1 skipped; 2,238 tests passed, 25 skipped; package check passed
- file-size, filesystem-boundary, typecheck, documentation, and diff checks passed
- Codex autoreview at P1 — clean
- fresh exact-head hosted checks and ClawSweeper review remain the landing gate

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Security and Correctness
 
+- Reject invalid runtime archive `onFiltered` policies before extraction instead of treating unknown values as entry-skipping opt-in.
 - Claim durable queue generations under a fail-closed cross-process lock with no-replace hardlinks and recoverable source retirement so acknowledgement and quarantine cannot delete or move a newer same-ID replacement; acknowledgement now requires a processing claim, and quarantine preserves existing failed-entry evidence on collisions.
 - Validate async/sync lock retry and deadline numbers and clamp synchronous backoff to the remaining finite deadline instead of overshooting or blocking forever.
 - Propagate asynchronous sidecar-lock release and failed-acquire deletion failures, preserve paired errors, and retain failed release state for safe retry through the same handle or manager drain.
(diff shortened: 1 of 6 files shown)


steipete:
@clawsweeper automerge

Special instructions:
Require the exact-head packed proof and hosted fallback/native checks before merge. Unknown runtime policy values must reject before destination mutation.

PR #172, opened 2026-08-30
fix(temp): pin sibling temp publication

steipete:
## Summary

- admit callback-produced sibling temps through one retained write-capable no-follow descriptor and exact bigint identity before any mode, sync, rename, or cleanup action
- reject symlink, directory/non-regular, hardlinked, oversized, unknown-identity, and replaced stages while preserving unverified callback paths and outside referents
- share the owner between `writeSiblingTempFile` and output-sibling publication, with descriptor mode/fsync, publication verification, and identity-bounded exit cleanup
- preserve producer-selected mode and disabled sync defaults when options are omitted, while applying explicit mode and opt-in file/parent sync through the retained owner

## Root cause

After the callback returned, sibling-temp publication trusted the generated pathname for chmod, reopen/sync, rename, and failure cleanup. A callback or concurrent actor could move the producer output and install a symlink, directory, FIFO, hardlink, or same-name replacement; later library operations could mutate, publish, block on, or delete that substitute.

Both callback-publication APIs now use one owner state machine: unadmitted pathname -> pre-open regular/single-link inspection -> retained read/write no-follow descriptor -> exact opened/current identity agreement -> requested descriptor mode and opt-in fsync -> serialized rename -> published-name identity verification -> close. No create or truncate flags are used during admission, and POSIX uses nonblocking open so FIFO substitution cannot hang the helper.

The final name is reverified through the retained descriptor after rename and after best-effort parent sync. A post-publication verification failure does not roll back or delete the final name. Failure cleanup only unlinks an admitted single-link regular file while parent, descriptor, and pathname identity still agree; observed replacements and all unadmitted callback output remain untouched and are removed from automatic cleanup authority.

## Live behavior proof

A fresh npm consumer installed the packed package and exercised the public `writeSiblingTempFile` API in native mode `off`:

```json
{"stable":{"result":{"filename":"stable.txt","token":7},"filePath":"stable.txt","content":"stable","mode":384,"events":["file-sync","rename"]},"producerModePreserved":{"content":"producer-mode","mode":416},"symlinkSubstitution":{"error":{"name":"FsSafeError","code":"symlink","message":"symlink sibling temp not allowed"},"tempIsSymlink":true,"outside":"outside","final":"old"},"directorySubstitution":{"error":{"name":"FsSafeError","code":"not-file","message":"sibling temp must be a regular file"},"retained":"keep","final":"old"}}
```

The explicit durability case synchronizes before rename, publishes at requested mode `0600`, and preserves the callback result. A second omitted-mode call preserves the producer's group-readable `0640`. Symlink and directory substitutions are rejected; both substitutes, the outside referent, and the previous final files remain unchanged.

## Verification

- focused sibling publication, compatibility, durability, platform, output, and stress suites: 7 files, 97 tests passed
- complete `CI=1 pnpm check`: 141 files passed, 1 skipped; 2,321 tests passed, 25 skipped; package/API validation passed
- build, docs examples, pack, file-size, filesystem-boundary, and diff checks passed
- fresh packed npm consumer proof passed
- Codex autoreview at P1: clean
- exact-head hosted Linux/macOS/Windows checks and ClawSweeper review remain the landing gate

## Compatibility and residual boundary

`writeSiblingTempFile` preserves historical defaults: omitted `mode` keeps the producer-selected permissions, and omitted sync options perform no file or parent sync. Explicit mode, `syncTempFile: true`, and `syncParentDir: true` use the retained owner; identity checks cannot be disabled. Default directory chmod remains best-effort only for the descriptor chmod call, while directory identity/open/type/close failures propagate. Callback failures and unadmitted outputs are intentionally preserved for caller-directed recovery instead of being deleted by pathname.

Identity checks and pathname rename/unlink remain separate syscalls. A hostile process can still change a leaf or parent in the final syscall gap or mutate bytes through another open descriptor. Use an approved writable directory plus cooperative locking or OS isolation when arbitrary concurrent namespace mutation is in scope. Parent sync remains best-effort rather than a strict durability receipt.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Security and Correctness
 
+- Pin callback-produced sibling temps through mode application, fsync, rename, and exact publication checks; reject symlink, non-regular, hardlinked, and replaced stages, preserve unverified cleanup paths, and default `writeSiblingTempFile` to mode `0600` with file and best-effort parent sync.
 - Make durable queue directory creation, claims, acknowledgement, quarantine, marker cleanup, and retirement transitions fsync affected directories and propagate durability failures.
 - Fsync synchronous file-store temps before rename and parent directories after publication, matching the documented durability contract.
 - Reject invalid runtime archive `onFiltered` policies before extraction instead of treating unknown values as entry-skipping opt-in.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Security and Correctness
 
-- Pin callback-produced sibling temps through mode application, fsync, rename, and exact publication checks; reject symlink, non-regular, hardlinked, and replaced stages, preserve unverified cleanup paths, and default `writeSiblingTempFile` to mode `0600` with file and best-effort parent sync.
+- Pin callback-produced sibling temps through requested mode application, opt-in fsync, rename, and exact publication checks; reject symlink, non-regular, hardlinked, and replaced stages and preserve unverified cleanup paths while retaining producer modes, disabled sync defaults, and best-effort directory chmod in `writeSiblingTempFile`.
 - Make durable queue directory creation, claims, acknowledgement, quarantine, marker cleanup, and retirement transitions fsync affected directories and propagate durability failures.
 - Fsync synchronous file-store temps before rename and parent directories after publication, matching the documented durability contract.
 - Reject invalid runtime archive `onFiltered` policies before extraction instead of treating unknown values as entry-skipping opt-in.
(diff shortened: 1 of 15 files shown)


steipete:
@clawsweeper automerge

Special instructions:
Require exact-head hosted proof in native-off and Windows lanes that callback symlink/directory/hardlink/FIFO substitutions are rejected without touching replacements or referents, staged mode/fsync occurs through one retained descriptor, and publication/cleanup replacement schedules never delete the observed substitute or roll back a post-rename final name.

steipete:
@clawsweeper automerge

Special instructions:
Compatibility defaults are restored: omitted mode preserves producer permissions, omitted sync options stay false, and only descriptor chmod failure remains best-effort for the directory. Verify exact-head hosted substitution/identity proof while retaining these upgrade behaviors.

PR #191, opened 2026-09-01
fix(atomic): preserve long destination filenames

steipete:
## What Problem This Solves

Fixes an issue where consumers writing or extracting valid long filenames would receive `ENAMETOOLONG`, or a wrapped filesystem error, because an internal temporary filename copied the entire destination basename before adding a random suffix.

On current main `790472f73bb499ec8eb0ee7a823de1e98b0a94c0`, this reproduces for 200/240-byte ZIP basenames, 240-byte `Root.write` and `Root.copyIn` basenames, and 240-byte synchronous JSON basenames. Each regression first creates and reads the destination directly to establish that the filesystem accepts it.

## Why This Change Was Made

Private staging names now use a fixed prefix and UUID independent of the destination. Archive extraction reuses its existing bounded default. JSON staging preserves literal parent segments and drive-relative semantics rather than normalizing the path through a different parent.

The repair leaves public producer/callback filenames, custom `tempPrefix`, final destinations, exports, configuration, permissions, and identity checks unchanged. The existing retained-descriptor, publication, and cleanup logic remains responsible for safety. A partial-write test now identifies write handles by their flags and asserts that a short write actually occurred, so it no longer depends on an internal filename containing the destination.

Production: **+10/−12, net −2**. Tests: **+59/−1, net +58**. Docs: **+12/−3, net +9**.

## User Impact

Consumers can use long filenames accepted by the filesystem without private staging consuming extra filename space. Short-name controls and the public naming contracts retain their behavior. No new fallback, setting, dependency, or migration is introduced. Release-note context is recorded here; changelog changes remain release-owned.

## Evidence

- On the same current-main checkout, the 11 selected regressions changed from **5 failed / 6 passed** to **11 passed**. The failures occur at private staging creation; the tests also verify replacement contents and absence of leaked temporary siblings.
- The 24 owner/security files passed **461 tests**, with **113 platform/native-dependent skips**.
- Canonical `pnpm check:changed` (the repository's `pnpm check` alias) passed lint, build, documentation checks, the full suite (**4,026 passed / 2,333 skipped**), and real root-tarball consumer installation/import/public-API checks. No dependency or lockfile changes and no test retries.
- Fresh independent managed P0–P2 review of the complete current twelve-file diff found no actionable findings.
- Native Windows execution was not performed locally. The separate native-platform `pnpm package:smoke` requires a prebuilt host `.node` payload that is absent locally; its normal CI jobs remain required before landing. The packed-consumer check above is not claimed as native-platform proof.

Focused command before and after:

```sh
pnpm test test/archive.test.ts test/fs-safe.test.ts test/json.test.ts \
  -t filesystem-admitted --reporter=json --outputFile=results.json
pnpm check:changed
```

Environment: macOS arm64, Node 24.20.0, pnpm 11.24.0. Introduction provenance is unknown.

### DIFF
--- a/docs/json.md
+++ b/docs/json.md
@@ -144,6 +144,9 @@ rename, so that fallback is temporarily non-atomic while retaining the staged
 file's `0600` mode; use the async `writeJson()`/`replaceFileAtomic()` surfaces
 when fallback policy must be explicit.
 
+Its private sibling temporary name is independent of the destination basename,
+so staging does not lengthen a valid destination filename.
+
 ```ts
 writeJsonSync("./prefs.json", { theme: "dark" });
 ```
--- a/docs/output.md
+++ b/docs/output.md
@@ -58,9 +58,12 @@ through the package's filename sanitizer; `fallbackFileName` supplies the name
 when nothing remains. This removes traversal, device-name, and invalid-character
 hazards but does not trim Windows-normalized trailing dots or spaces; reject or
 rewrite those when cross-platform filename uniqueness matters.
-The same sanitized basename is used for producer staging, guarded internal
-temps, the final rename target, and the returned `path`; raw and staged names
-never diverge.
+`staging: "workspace"` passes the sanitized basename to the producer.
+`staging: "sibling"` embeds that basename in its randomized temporary name.
+The final target and returned `path` use the destination basename, sanitized
+when needed as described above. Guarded temporary files used only inside
+fs-safe have independent names so their length does not grow with the
+destination basename.
 
 ## Choosing a staging mode
 
--- a/docs/writing.md
+++ b/docs/writing.md
@@ -27,6 +27,9 @@ await fs.mkdir("snapshots/2026/05");
 5. Atomically rename the temp file over the destination.
 6. Stat the resulting fd and verify identity.
 
+Private sibling temporary names are independent of the destination basename,
+so staging does not add a suffix to an otherwise valid long filename.
+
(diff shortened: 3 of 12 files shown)


steipete:
Merged as [8d4ff13365c33db031c4b7cbd48e40777a98733c](https://github.com/openclaw/fs-safe/commit/8d4ff13365c33db031c4b7cbd48e40777a98733c). All twelve changed files match the reviewed head, and the task checkout is clean on main after a fast-forward pull. A fresh managed P0–P2 review found no actionable findings. Production: **net −2 lines**; tests: **+58**; docs: **+9**. No package version, tag, or npm publication was performed.

The long-filename repair retains the guarded publication and identity checks. All eleven selected regressions pass after five intended baseline failures; normal `pnpm check:changed` passed lint, build, docs, 4,026 tests, and the packed-consumer API check. [CI at the exact PR head](https://github.com/openclaw/fs-safe/actions/runs/33500175395) also passed the Node 22/24 matrix, Linux/macOS/Windows native checks, musl check, and bundled package smoke jobs. The new long-name tests run through their explicit fallback mode; this does not claim an additional long-name native Windows test that was not run.

The ClawSweeper changelog finding and matching rank-up move are intentionally skipped: changelog/version edits belong to release preparation. This repair PR preserves the release-note context in its body, and does not choose a future version, tag, or publish a package. The user-visible change is that filesystem-admitted destination basenames no longer overflow when private atomic-write/archive staging names are created. No remaining code or security finding was reported.

PR #214, opened 2026-09-03
fix(queue): propagate batch transition failures

steipete:
## Summary

`loadPendingJsonDurableQueueEntries()` used one broad per-entry catch around claim, read/parse, caller migration, and migration publication. That intentionally skipped malformed entry content, but it also suppressed queue-owned transition failures: post-link claim directory sync and migration write/fsync/publication errors became `[]` or partial success even though namespace state had changed or durability was unconfirmed.

The batch loader now keeps only entry read/identity/size/JSON parsing and the caller `read` callback inside the skippable boundary. Claim, transfer-lock, retirement, and migration publication failures propagate with the original error. Initial unowned hardlink/unknown-identity admission remains an explicit internal `skipUnowned` claim result; post-link mismatches still reject.

Migrated-entry publication now omits the generic best-effort `syncParentDir` option and performs queue-owned strict `syncDirectory()` after publication. Both single and batch loaders therefore reject real post-rename directory-sync failures while preserving already-visible migrated bytes for safe retry.

Production delta is net +6 lines. The main queue source remains within its 500-line limit at 497 lines. Focused regressions add 324 lines across 17 cases. No public signature, declaration, export, native ABI, dependency, lockfile, generated output, timeout, or budget changed. Acknowledgement behavior (task #96) and enqueue parent sync (task #100) are explicitly unchanged.

## Physical syscall proof

Baseline Linux arm64 packed-main/published evidence used exact Node 24.20.0, native `off`, `strace` 6.19, and ptrace syscall injection:

```text
batch claim fsync EIO      -> returned []; json + processing remained same inode/nlink 2
single claim fsync EIO     -> rejected EIO; same recovery state
batch migration temp EIO   -> returned []; old processing bytes remained
single migration temp EIO  -> rejected EIO; old processing bytes remained
```

Every retry and fault-free control succeeded.

A fresh physical install of the packed candidate was then exercised in a Linux arm64 local container with exact Node 24.20.0 and `strace` 6.1. Eight control/fault cases passed:

- Batch and single claim both rejected injected directory `fsync` EIO; both preserved `.json` and `.processing` as same-inode two-link recovery evidence; retry succeeded.
- Batch and single migration both rejected injected temp-file `fsync` EIO; old processing bytes remained; retry published migrated bytes.
- Batch and single migration both rejected injected post-publication directory `fsync` EIO; migrated bytes were already visible; retry successfully republished/resynced.
- Fault-free batch claim and migration controls succeeded.
- Every injected trace recorded exactly one kernel `fsync(...)= -1 EIO`; every retry returned the expected entry.

Candidate root artifact integrity: `sha512-ne1nhtvDyAQDhPuvgZJDAlJszlEXPtC2fTh9f6RUupS+S3PqEQKK0GhXtIViZ9rMF1DhySHKnP+YU+o2YELT0A==`.

This is real syscall-boundary ptrace fault injection, not physical-media or crash-loss proof. Linux evidence is local-container proof, not remote/AWS proof.

## Adversarial coverage

The new suite verifies:

- First and second claim-directory sync failures propagate and retain hardlink recovery state.
- Transfer-lock acquisition and post-link path/extra-hardlink validation failures propagate.
- A failure after an earlier batch entry succeeds rejects rather than returning a misleading partial array, while both claims remain retryable.
- Migration write, temp sync, rename, post-publication verification, and strict directory sync failures retain the correct old/already-visible generation and retry safely.
- Initial hardlinked and unverifiable pending entries still skip explicitly; single-loader controls continue to reject them.
- Read identity failures, malformed/oversized content, callback errors, symlinks, non-files, and missing inputs retain their established batch skip behavior.
- Mixed valid entries retain listing order and result shape.

## Validation

- Focused queue durability/generation/failure/security suites: 75 passed.
- New batch durability suite independently on exact Node 22.0.0, 22.23.2, and 24.20.0: 17 passed on each runtime.
- `CI=1 pnpm check`: 6,915 passed / 80 skipped; 200 files passed / 2 skipped.
- `pnpm test:security`: 84 passed.
- `pnpm package:smoke`: npm/pnpm physical consumers, host native loading, optional omission, and missing-binary behavior passed.
- File-size, filesystem-boundary, build, docs, package, and `git diff --check` gates passed.
- Independent Codex autoreview was scoped-clean at P0 (0.98).

Exact-head cross-platform CI and ClawSweeper review are required before merge. No release or package publication is included.


steipete:
Addressed the P1 malformed-ID regression in `b4671e3f62272b8ce88c0f156e622826213c0498`.

Invalid IDs derived from untrusted directory entries are now filtered before path resolution or claim. This restores the prior skip behavior for `.json`, `.processing`, dot-prefixed IDs, and names containing spaces while leaving claim, transfer-lock, retirement, and migration transition failures outside the skippable boundary.

The new regression places all four malformed names before a valid `job.json`, verifies the valid entry is returned, and verifies every malformed file remains byte-exact and untouched. Focused queue suites pass 76/76, the queue source remains within budget at 498 lines, `git diff --check` passes, and the complete updated diff is autoreview-clean at P0 (0.97).

@clawsweeper re-review

steipete:
Landed as `7e3350b2508908d378a8ec932bd3664f42cb4d62` with tree `bcfb9752ef0f333499e179d6ad47b091c1b573b9`.

Verified final head `b4671e3f62272b8ce88c0f156e622826213c0498` before merge:

- focused queue suites: 76/76 passed;
- full check: 6,915 passed / 80 skipped;
- security suite: 84 passed;
- bundled package smoke and `git diff --check`: passed;
- exact-head CI: Node 22/24 on Linux, macOS, and Windows; native Linux/macOS/Windows/musl; package smoke on all three OSes; coverage and merged coverage; CodeQL; Cargo clippy/audit; workflow lint; and benchmarks all passed;
- complete-diff autoreview: no accepted/actionable P0 finding (0.97);
- replacement ClawSweeper review: no actionable correctness or security finding, with physical proof sufficient.

Physical proof used a fresh packed candidate in a Linux arm64 local container with exact Node.js 24.20.0 and `strace` 6.1. Eight control/fault cases exercised batch and single-entry claim, migration temp-file sync, and migration post-publication directory sync. Each injected case contained exactly one kernel `fsync(...)= -1 EIO`; the call rejected with the original error, retained the expected retry state, and succeeded on retry. This is syscall-boundary fault injection, not physical-media or crash-loss proof.

The final repair preserves narrow skipping for malformed/unreadable input while making queue-owned claim, retirement, transfer-lock, and migration durability failures reject for caller retry. The review follow-up also restores malformed listed-name skipping before claim, so attacker-created `.json`, `.processing`, dot-prefixed, or space-containing names remain untouched and cannot block later valid work.

PR #215, opened 2026-09-03
fix(queue): resync completed acknowledgement retries

steipete:
## Summary

- resync the durable queue directory when acknowledgement is retried after the final `.delivered` unlink is already visible;
- perform that recovery sync before either reporting completion or rejecting a newer pending generation;
- add focused terminal-state regressions, document the retry contract, and record the public operational fix under 0.7.3 Unreleased.

## Root cause

Acknowledgement correctly performs:

1. rename `.processing` to `.delivered`;
2. sync the queue directory;
3. unlink `.delivered`;
4. sync the queue directory again.

If step 4 fails, the call rejects but both marker names are already absent. The old retry path treated that all-absent state as complete and returned without another sync. The final unlink could therefore remain unconfirmed even after the caller followed the documented retry contract. If a newer `.json` generation was waiting, retry returned the generation-bound `path-mismatch` before repairing the prior acknowledgement sync.

The filesystem cannot distinguish a fully durable all-absent state from a visible unlink whose directory sync failed. The narrow repair is therefore an idempotent strict queue-directory sync under the existing per-entry transfer lock whenever `.processing` is absent, after optional delivered-marker cleanup and before inspecting `.json`.

## Compatibility and boundaries

This intentionally adds one directory sync to already-completed acknowledgement calls while the queue directory exists. Real sync or identity failures continue to propagate, and a waiting newer generation still produces the same `path-mismatch` after recovery succeeds.

The change preserves queue-directory-absent idempotency, rename-before-unlink ordering, generation ownership, transfer-lock serialization, and platform-specific directory-sync support policy. It does not change batch loading, migration, enqueue publication, delivered-marker batch cleanup, exports, or types.

Production delta: one line moved, net zero lines.

## Validation

- focused terminal acknowledgement regressions: 2 passed on exact Node.js 22.0.0, 22.23.2, and 24.20.0;
- queue ownership regression set: 77 passed;
- full `CI=1 pnpm check`: 200 test files passed / 2 skipped, 6,918 tests passed / 80 skipped;
- security suite: 84 passed;
- release-package smoke: passed;
- `git diff --check`: passed;
- Codex autoreview: no accepted/actionable findings, patch correct at 0.99.

## Physical syscall-boundary proof

A fresh packed candidate was installed in a Linux arm64 local container using exact Node.js 24.20.0, native mode `off`, no loaded package `.node` binding, and `strace` 6.1. Candidate package SRI:

`sha512-kgaeLvZuuyDYPkhSGDTKyI6ZBVtcEAgR6RsXAvyGb2q5Eo+H2XrlwlpIrm3MsO715wh7kEnO/uqCa3PR6OcEfQ==`

Eleven setup/control/fault observations passed their assertions:

- control acknowledgement issued two queue-directory `fsync` calls; a completed retry issued one additional successful `fsync`;
- faulted acknowledgement showed `rename(.processing -> .delivered)`, successful directory `fsync`, `unlink(.delivered)`, then exactly one injected `fsync(...)= -1 EIO`;
- in a fresh retry process with both markers absent, injection at its first `fsync` produced exactly one `EIO`, proving retry reached and propagated the recovery barrier;
- a final healthy retry issued one successful `fsync` and completed;
- with generation 2 waiting at `.json`, both injected failures preserved its exact bytes; the final healthy retry synced first, then returned the existing `path-mismatch` and still preserved generation 2.

The earlier packed/released behavior was also captured at the actual Node/libuv syscall boundary: after the same second-`fsync` failure, retry returned with no further `fsync` in the trace.

This is local-container ptrace syscall fault injection. It proves actual kernel syscall ordering, error propagation, and retry-visible filesystem state; it is not physical-media or crash-loss proof.


steipete:
Landed as `e940b0da871733713ebc1ddb16ab11366002d4d3` with tree `dfe78e763bb60862835d496fa75d12e078659eb0`.

Verified final head `eb340f42e4e4ac6f71a484b12f9f7fc0a1c0704f` before merge:

- focused terminal-state regressions: 2/2 passed on exact Node.js 22.0.0, 22.23.2, and 24.20.0;
- queue ownership regression set: 77 passed;
- full check: 200 test files passed / 2 skipped, 6,918 tests passed / 80 skipped;
- security suite: 84 passed;
- release-package smoke and `git diff --check`: passed;
- Codex autoreview: no accepted/actionable finding, patch correct at 0.99;
- ClawSweeper: no correctness or security finding, with physical proof sufficient;
- exact-head CI: Node 22/24 on Linux, macOS, and Windows; native Linux/macOS/Windows/musl; package smoke on all three OSes; coverage and merged coverage; CodeQL; Cargo clippy/audit; workflow lint; and benchmarks passed.

Physical proof installed a fresh packed candidate in a Linux arm64 local container with exact Node.js 24.20.0 and `strace` 6.1. The trace showed `.processing -> .delivered`, the first successful directory `fsync`, `.delivered` unlink, and an injected `EIO` on the second `fsync`. A fresh retry with both markers absent reached the repaired directory-sync barrier and propagated a separately injected `EIO`; a final healthy retry synced and completed. With generation 2 waiting at `.json`, both failures preserved its exact bytes and the final healthy retry synced before returning the existing `path-mismatch`. This is syscall-boundary fault injection, not physical-media or crash-loss proof.

The first Windows Node 22 CI attempt failed outside this PR’s queue surface when a PowerShell ACL-owner probe timed out and a sidecar process-exit test also timed out with cleanup residue. The exact failed job passed unchanged on rerun; no acknowledgement test failed.

The production repair is one moved line (+1/-1, net zero): an all-absent acknowledgement retry now uses the existing strict guarded queue-directory sync before reporting completion or inspecting a newer pending generation.

PR #225, opened 2026-09-04
docs(write): scope create-only publication visibility by backend

steipete:
## Summary

Live testing of `main` confirmed a real contract violation: in the pure-JavaScript fallback, `Root.create()`, `createJson()`, and `write({ overwrite: false })` claim the final destination pathname with `O_EXCL` before content is written, so a concurrent observer can see an empty file at the target. The [writing docs](docs/writing.md) promised sibling-temp commit with no half-written replacement visible for these operations.

A staging fix was implemented and fully tested locally (private temp file published via hardlink link+unlink). It worked, but the hosted sidecar-contention proof caught a worse regression: between the link and the unlink, the published name transiently has two links, and concurrent readers/creators that enforce single-link identity fail with `hardlink`/`path-alias` errors instead of clean contention. POSIX plus Node's fs API offers no atomic operation that is simultaneously no-clobber, content-complete, and single-link: `rename` clobbers, `O_EXCL` open exposes the name early, and link+unlink exposes a two-link window. The native binding's `renameNoReplace` is the correct mechanism and already provides the guarantee.

So this PR resolves the defect at the contract boundary rather than by regressing concurrent behavior:

- **Docs**: the sibling-temp no-visibility guarantee for create-only writes is scoped to backends with atomic no-replace publication (the native binding, default in `auto`/`require`). The pure-JavaScript fallback's claim-then-write behavior is documented explicitly, matching the package's existing best-effort fallback model.
- **Regression tests**: new coverage pins both behaviors — the fallback claims the name before content (so a future change is deliberate), and the native backend keeps the destination absent until publication. Outcome semantics (final bytes, `already-exists` collision preservation, failure cleanup) are asserted for all four create-only operations.
- **Changelog**: records the contract clarification.

## Evidence

- Defect reproduced on installed consumers of `main` with a separate-process observer: all four create-only operations exposed the destination at size 0 before content, in native-off mode; the same probe in require mode saw the destination absent until publication.
- The intermediate hardlink implementation passed all local suites but failed hosted macOS Node 24 CI (`sidecar-contention-proof` root-async: `hardlink`/`path-alias` errors, 30/400 acquisitions before the watchdog). That failure is the reason the code change was withdrawn; it is described here rather than hidden.
- After reverting to the documented-behavior resolution: full CI-bounded check, security suite, and both sidecar-contention modes pass locally; hosted CI must pass before merge.

## Boundaries

No runtime code changes ship in this PR. No API changes, no new options. The fallback limitation is a Node/POSIX capability gap, not a new relaxation: native (the default) already satisfies the strong contract.


steipete:
Addressing the review on head b967e9d5 (docs precision fix pushed).

**1. Fallback staging claim corrected.** The docs no longer say create-only writes "share that staging." They now state explicitly that sibling-temp staging with atomic no-replace publication exists only on backends that provide it, and that the pure-JavaScript fallback does not stage: it claims the final name with `O_EXCL` and writes content in place. The `fs.create` section carries the same distinction.

**2. Real-behavior evidence.** Since this PR changes no runtime code, the relevant proof is the documented current behavior, executed on installed packages from actual main (`027f643`, the runtime-ancestor of this branch; the two later main commits touch CI fixtures and dev dependencies only). A separate Node process observed the real destination while each create-only call was paused immediately before its first content write — no mocked filesystem state:

```json
{
  "probe": "separate-process observer at first content write",
  "nativeOff": [
    {
      "operation": "create",
      "beforeWriteExists": true,
      "beforeWriteBytes": 0,
      "finalContentVerified": true
    },
    {
      "operation": "write-no-replace",
      "beforeWriteExists": true,
      "beforeWriteBytes": 0,
      "finalContentVerified": true
    },
    {
      "operation": "createJson",
      "beforeWriteExists": true,
      "beforeWriteBytes": 0,
      "finalContentVerified": true
    },
    {
      "operation": "writeJson-no-replace",
      "beforeWriteExists": true,
      "beforeWriteBytes": 0,
      "finalContentVerified": true
    }
  ],
  "nativeRequire": [
    {
      "operation": "create",
      "beforeWriteExists": false,
      "finalContentVerified": true
    },
    {
      "operation": "write-no-replace",
      "beforeWriteExists": false,
      "finalContentVerified": true
    },
    {
      "operation": "createJson",
      "beforeWriteExists": false,
      "finalContentVerified": true
    },
    {
      "operation": "writeJson-no-replace",
      "beforeWriteExists": false,
      "finalContentVerified": true
    }
  ]
}
```

In native-off mode the destination existed at size 0 before content (the documented limitation now stated in the docs). In require mode the destination was absent until publication (the guarantee). The regression tests added in this PR execute the same assertions in the repo: fallback name-claim visibility is pinned at `test/root-create-publication.test.ts`, and the native no-visibility guarantee is asserted through the real bundled binding on all CI platforms (skipped only where no binding exists).

The withdrawn hardlink implementation's failure is preserved in the PR description: it passed all local suites but regressed hosted sidecar contention through its transient two-link publication window, which is why the code change was abandoned rather than weakened further.

@clawsweeper re-review

steipete:
Addressing the re-review finding on head 1502091 (pushed).

**Default `auto` mode claim qualified.** The docs previously said the no-visibility guarantee comes with "the native binding used by the default `auto` and `require` modes." Corrected: `auto` only uses the native binding when it loads successfully, so the guarantee is now stated as `require` mode, or `auto` mode with a successfully loaded binding, and callers who need the guarantee are directed to `require` mode. The `fs.create` section uses the same qualification.

No code changes; the observer receipts in the previous comment remain the behavior evidence.

@clawsweeper re-review

steipete:
Merged as [`e875fdc`](https://github.com/openclaw/fs-safe/commit/e875fdc718ba8bb128a5efe29b1d53b45f756059) and verified on clean `main` (all 20 new contract tests pass there).

**How this resolved.** Live testing with a separate-process observer confirmed the original contract violation on `main`: in native-off mode, `create`, `createJson`, and both `overwrite: false` forms exposed the final pathname at size 0 before content was written. A staging fix (private temp + hardlink link+unlink publication) was implemented and passed every local suite, but hosted CI caught the catch: the link→unlink sequence leaves a transient two-link window, and the macOS sidecar-contention proof failed with `hardlink`/`path-alias` errors from concurrent readers enforcing single-link identity. POSIX + Node's fs API offers no operation that is simultaneously no-clobber, content-complete, and single-link, so the writer change was withdrawn rather than weakening the hardlink-rejection contract.

The landed change resolves the defect at the contract boundary: the no-visibility guarantee is scoped to backends with atomic no-replace publication (`require` mode, or `auto` with a successfully loaded binding), the fallback's claim-then-write behavior is documented explicitly, and regression tests pin both behaviors so any future backend change is deliberate. Final review confirmed all three earlier blockers resolved and endorsed the docs-only approach as safer than the withdrawn implementation.

**Verification.** Full CI and coverage green on the reviewed head `1502091` (including Windows, macOS, Linux, musl, and the sidecar-contention proofs in off and require modes); security suite 84/84 locally; the flaky dispatch-fixture timeout and the cancelled coverage job were rerun to green. One pre-existing Windows stream-limit cleanup assertion flaked in a pre-rebase run of unchanged code and passed on rerun.

No release or publication was performed.

PR #227, opened 2026-09-04
feat(file-lock): add retainOnExit opt-out for process-exit cleanup

steipete:
## Summary

0.8.0 started releasing Root-backed sidecar locks on natural event-loop shutdown. That is the right default, but it broke a legitimate consumer pattern: OpenClaw's dist-artifact ownership deliberately retains its lock file across exit when a build child leaves unjoined work, because PID death alone cannot prove the writer stopped.

This adds `retainOnExit?: boolean` to `acquireFileLock`/`withFileLock` options. When set, the lock's sidecar is exempt from both the `beforeExit` Root cleanup and the synchronous `exit` fallback, and remains governed only by the caller's stale policy. Default behavior is unchanged.

## Validation

- New child-process tests prove an unchanged sidecar acquired with `retainOnExit: true` survives natural exit in both Root-backed and raw modes; existing tests continue to prove default release, changed-sidecar preservation, and single-attempt failure cleanup.
- Full gate: 7,512 passed / 80 skipped. The two `consumer-pnpm-lifecycle` failures reproduce identically on unmodified main in this environment (the suite requires a pnpm lifecycle context this shell does not provide); they are unrelated to this change.
- Docs (`docs/sidecar-lock.md`) and changelog updated.

## Boundaries

Explicit `release()` still works normally; the option only exempts the process-exit handlers. This is an opt-out for fail-closed ownership records, not a way to skip cleanup silently.


steipete:
Addressing the review on head 54f875c (pushed).

**P2 — reset() independence:** correct; `releaseAllLocksSync` is shared by the exit handler and `FileLockManager.reset()`. The exit handler now passes `preserveRetained: true`; explicit `reset()` retains its teardown semantics and removes retained raw locks. Covered by a new in-process regression (`manager reset still releases a retainOnExit raw lock`).

**P3 — documented option shape:** `retainOnExit?: boolean` is now listed in the acquire-options block in `docs/sidecar-lock.md`, and the exit-behavior section notes the exemption.

**Real-behavior proof:** direct child-process runs against built `dist` (not the test harness), each acquiring a Root-backed lock then exiting naturally:

```json
{
  "default": { "lockFileSurvivesExit": false },
  "retainOnExit": { "lockFileSurvivesExit": true }
}
```

The default mode removed the sidecar at exit; the retained mode preserved it. The same proof passes on the child-process suite in CI. Full serial suite: 7,513 passed / 80 skipped; the two `consumer-pnpm-lifecycle` failures reproduce identically on unmodified `main` in this shell (missing pnpm lifecycle context) and are unrelated.

@clawsweeper re-review

steipete:
Addressing the re-review on head 70ea15a (pushed).

**P1 — mixed-version exit handlers:** agree, and fixed fail-closed. Exit-handler registration now stamps a retain-aware marker (`fsSafe.sidecarLockRetainAwareCleanup`) alongside the existing dedup keys; when the cleanup handler is registered but the marker is absent (an older copy owns it), a `retainOnExit` acquisition rejects with `helper-unavailable` before touching the filesystem. A retained sidecar can never be silently deleted by a legacy handler. A regression test simulates the legacy-registered state (global dedup keys set, marker absent) and asserts the refusal; normal acquisition is unaffected.

The same reasoning covers the Root-backed `beforeExit` path: that registration is also keyed and deduplicated globally, and the same marker gates it.

**Verification:** the exit suite passes 15/15 (retained raw + Root survival, reset independence, legacy refusal, plus the existing default-release and changed-sidecar cases). Full gate: 7,514 passed / 80 skipped; the two `consumer-pnpm-lifecycle` failures reproduce on unmodified `main` in this shell and are unrelated. The dist-level child-process proof from the previous comment still holds: default exits remove the sidecar, retained exits keep it.

@clawsweeper re-review

steipete:
Addressing the re-review on head 0b761f6 (pushed).

**P2 — reentrant upgrade:** agree. Retention is now monotonic: a same-owner acquisition with `retainOnExit: true` upgrades the held entry (`held.retainOnExit = true`) instead of silently reusing the default policy, and a later default acquisition never downgrades it. Two regressions pin this: an in-process check that the held entry's flag upgrades under a doubled refCount, and a child-process natural-exit run for the default-then-retain sequence (sidecar survives, payload intact).

Re-verified against built `dist`:

```json
{
  "default": { "lockFileSurvivesExit": false },
  "retainOnExit": { "lockFileSurvivesExit": true }
}
```

Serial suite: 7,516 passed / 80 skipped; the two `consumer-pnpm-lifecycle` failures reproduce on unmodified `main` in this shell (missing pnpm lifecycle context) and are unrelated.

@clawsweeper re-review

steipete:
Maintainer disposition for the remaining review item: **`retainOnExit` is accepted as a supported public contract.**

Rationale: the package's documented model is that natural-exit cleanup is best-effort and callers with fail-closed ownership records already exist (OpenClaw's dist-artifact build lock, where PID death cannot prove a detached writer stopped). The option is opt-in, defaults are unchanged, mixed-version processes fail closed rather than silently losing the guarantee, and stale recovery remains the caller's explicit policy. Documented in `docs/sidecar-lock.md` with the retention and failure modes.

The [current-head review](https://github.com/openclaw/fs-safe/pull/227#issuecomment-5546293363) (2026-09-04T21:49:07Z, head `0b761f6a`) reports no actionable code or security findings; its `needs-human` marker is not being treated as approval — this comment is the maintainer judgment it asked for.

All exact-head CI, coverage, benchmarks, and CodeQL gates pass. Serial local suite: 7,516 passed / 80 skipped; the two `consumer-pnpm-lifecycle` failures reproduce on unmodified `main` in this shell and are unrelated.

PR #231, opened 2026-09-05
fix(atomic): tolerate FAT-family rename identity drift on exFAT/FAT32

dongsheng123132:
## Summary

Windows mints a fresh file identity (new ino) on rename when names exceed the 8.3 short-name budget. Observed on real hardware: destination basenames >=16 chars get a fresh ino after rename on exFAT/FAT32 USB sticks, <=15 chars keep it. replaceFileAtomic then fails with path-mismatch even though the bytes landed correctly -- every atomic write to a USB stick with a normal-length filename fails (e.g. OpenClaw gateway config writes, portable USB apps).

## Changes

- Temp/staging basenames stay within 8.3 (<=12 chars): buildReplaceTempPath, sibling-temp, output-sibling, json temp names, all now xxxxxxxx.tmp.
- Post-rename verification re-anchors to the published file's live identity after structural checks (symlink/hardlink/not-file rejection, fresh descriptor-vs-pathname agreement) PLUS a staged-bytes equality gate. Swapped bytes still fail closed with path-mismatch; never deletes or rolls back the published file.
- mkdir -p skipped when the directory exists (FAT USB roots reject mkdir-with-mode with EPERM even for existing dirs).
- Hostile tempPrefix (path separators) still throws instead of silently collapsing.

## Test plan

- [x] pnpm build clean
- [x] pnpm test: 185 files passed, 4307 tests passed (38 files / 3187 tests skipped, same as main)
- [x] Real-hardware repro: exFAT + FAT32 + NTFS USB sticks, 5/5 pass (async/sync replace, dest basenames 12-18 chars 7/7 pass)
- [x] New regression tests: content-identical copy+unlink publication resolves; swapped-bytes-after-rename still rejects (async + sync)
- [ ] pnpm check file-size budget: 3 files exceed 500 lines (temp-owner 538, replace-file 512, callback-substitution 526) -- left for maintainer to decide on budget vs split; all other check steps pass

## Repro (minimal, Windows + exFAT USB stick)

replaceFileAtomic({ filePath: 'E:/state/xxxxxxxxxxxxxxxx.json', content: 'new' })
before: rejects path-mismatch (bytes correct on disk). after: resolves rename.

Environment: @openclaw/fs-safe 0.8.1, openclaw 2026.9.1, Node v22.22.3, Windows 11, exFAT + FAT32 USB sticks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Fix atomic writes on FAT-family filesystems (exFAT/FAT32 USB sticks): Windows mints a fresh file identity on rename when names exceed the 8.3 short-name budget, so `replaceFileAtomic` failed with `path-mismatch` even though the bytes landed. Temp/staging basenames now stay within 8.3 (<=12 chars); post-rename verification re-anchors to the published file's live identity after structural checks plus a staged-bytes equality gate (swapped bytes still fail); `mkdir -p` is skipped when the directory exists (FAT USB roots reject mkdir-with-mode).
 - Fix `replaceFileAtomic({ dirMode })` rejecting a raw `fs.stat` mode: directory modes are masked to permission bits (`0o7777`) before application and verification, matching chmod semantics; 0.8.0 regressed this input tolerance with `directory final mode could not be verified`.
 
 ## 0.8.1 - 2026-09-04
--- a/src/json.ts
+++ b/src/json.ts
@@ -187,7 +187,10 @@ export function tryReadJsonSync<T = unknown>(
 
 export function writeJsonSync(pathname: string, data: unknown) {
   // Keep literal parent segments so staging follows the same symlinks as the target.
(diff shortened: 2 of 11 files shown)


steipete:
Closing this implementation after independent review with real filesystem probes.

The exFAT/FAT32 rename identity drift you report is a real problem, and thank you for the hardware evidence. The proposed fix, however, changes the security contract for every caller: with the branch's changed modules, explicit strict mode accepted a substituted inode containing identical bytes with mode 0777 (main rejects it with path-mismatch), and tempPrefix "save" consumed an unrelated existing save.tmp (main preserves it). The branch also replaces the shared prefix validation and exceeds the enforced file-size limits.

Main already offers an explicit opt-in for this situation via renameIdentity: "verify-content-with-lock", with documented limitations. Transparent acceptance would weaken existing callers without their opting in, so we will not land it in this form. Your report remains useful evidence for future work under that explicit compatibility policy.

PR #357, opened 2026-09-14
fix(security): make no-clobber root moves atomic

steipete:
Root moves without overwrite could race a separate destination check and replace a file created by another writer. This change uses native atomic no-replace rename, snapshots overwrite before admission, and rechecks deny rules against the canonical source and destination actually selected by native parent admission. Missing atomic support fails closed with the documented helper-unavailable result.

Native-off isolated producers retain a zero-copy hard-link handoff. Maintainer review found and fixed legacy Windows deletion behavior: the pin transfers to an already-verified sibling descriptor before source retirement, and a cleared read-only attribute is restored and verified through that descriptor. Identity, one-link publication, original callback receivers, cleanup ownership, and existing sync access requirements remain enforced.

The integrated JavaScript suites pass 80 tests (2 platform skips), all 88 Rust workspace tests pass, and fresh independent review through P2 is clean. Real Windows NTFS probes reproduced the original directory-cleanup and read-only-bit failures on Node 22.0.0 and proved the descriptor-transfer/restoration approach; current Node 24 controls also passed. The packed package passed real Windows NTFS tests on Node 22.0.0 and Node 24.11.1: identity, contents, mode, and cleanup were verified, with existing write-access failures preserved for read-only outputs requesting sync. A follow-up preserves the established Windows admission-error precedence when cleanup also fails; 23 focused tests pass and independent review is clean. Final head `c6e02f3c47f5970cd3d5abbf8b9c3dd32a9ab944` passed the full CI matrix, including Windows native coverage of admission-error precedence, and was fast-forwarded onto main.


steipete:
Exact-head runtime evidence for `1abdf368a16309999bc078919b058771f911171b` on Windows x64:

```text
Node 22.23.2
root-move-native-integration.test.ts
{"proof":"root-move-native-no-replace-race","competitor":"preserved","source":"preserved"}
1 passed

sibling-producer-isolation.test.ts
temp isolated sibling producer > publishes native-off output without copying or changing its identity
output isolated sibling producer > publishes native-off output without copying or changing its identity
2 passed
```

The native collision case loaded the real bundled addon through the production loader, used mode `require`, created the competitor after native parent admission, and observed typed `already-exists` with both source and competitor bytes intact. The isolated producer cases used mode `off`, real filesystem hard links, exact inode preservation, final `nlink === 1`, expected bytes, and no leftover sibling stage. The same native collision proof also passed on Node 26.8.2. The full GitHub Linux/macOS/Windows matrix is green.

Owner decision: accept the documented upgrade boundary. Public default no-clobber `Root.move()` is native-required; restoring a check-then-replacing-rename fallback is not acceptable. Native-off isolated producer publication requires hard-link support and fails explicitly on unsupported filesystems rather than copying or replacing a competing stage.

Final performance qualification is still pending and will be posted separately before merge.

@clawsweeper re-review

PR #368, opened 2026-09-14
fix(security): keep Windows fds in the host runtime

steipete:
## Summary

- keep Windows Node/Bun descriptors in the host runtime that created them
- resolve a paired `uv_get_osfhandle` / `uv_open_osfhandle` bridge and fail closed when the pair is missing or incomplete
- keep raw-HANDLE operations internal, preserve descriptor ownership, and reject invalid descriptors
- extend the Bun proof and add focused bridge regression coverage

## Security and compatibility policy

The native addon previously interpreted a Node/Bun file descriptor through the addon's own Windows CRT. CRT descriptor tables are runtime-local, so the same integer could be invalid or refer to the wrong object. Native operations could therefore lose the descriptor authority established by the JavaScript caller.

Descriptors now convert only through the originating host runtime. Required native descriptor operations return `ENOTSUP` when the paired bridge is unavailable; namespace guessing and the foreign-CRT fallback are intentionally not retained. `auto` mode continues to use existing guarded portable fallbacks where that API has one, while `require` remains fail-closed. Node 22/26 and Windows Bun 1.4.2 are qualified below.

## Validation

- exact rebased head `1943799`: `pnpm native:build` and `pnpm build` passed
- `pnpm native:test`: archive core 34 passed; native addon 39 passed
- focused bridge/native tests on Node 22.23.2: 67 passed, 17 skipped
- focused bridge/native tests on Node 26.8.2: 67 passed, 17 skipped
- package build, docs check, file-size boundary check, fs-boundary check, and pack check passed
- Astra independently challenged the original implementation, rebased integration, performance evidence, and Bun proof with no P0-P2 blockers

## Windows Bun 1.4.2 proof

Exact `bun --jitless scripts/bun-native-proof.mjs` run on Windows x64 exited 0 for every scenario:

| Scenario | Result |
| --- | --- |
| `auto` | native addon loaded; Root confinement and native-policy SHA passed |
| `require` | borrowed root fd remained usable; native-returned fd was read, bigint-statted, and closed directly by Bun |
| `off` | ordinary guarded operations passed; zero addon loads |
| missing addon + `auto` | portable operations and hashing passed; zero addon loads |
| missing addon + `require` | hashing failed `helper-unavailable`; input and directory entries were preserved |

Built addon SHA-256: `7056c6a2306012308aa2dec4ed802002a8c04aea48892b7b9c5819a7c7c364f3`.

This is the production-path descriptor and mode-compatibility proof, not a claim that the complete Bun Vitest suite was run locally. Failed-conversion cleanup and invalid-fd behavior are covered by focused Rust/Vitest tests.

## Performance

Windows x64, exact measured baseline `aa677a8` and candidate `4d537b6`, 30 balanced pairs per case, every measured batch >=250 ms, 10,000-sample bootstrap. The rebased `1943799` native tree and benchmark bodies are byte-identical to the measured candidate. Acceptance required the upper 95% ratio bound to remain <=1.05.

| Runtime | Case | Candidate / baseline median | 95% interval |
| --- | --- | ---: | ---: |
| Node 22 | `fstatIdentity` | 0.9459 | 0.9436-0.9490 |
| Node 22 | `openBeneath` + close | 0.9849 | 0.9759-0.9922 |
| Node 26 | `fstatIdentity` | 0.9406 | 0.9384-0.9440 |
| Node 26 | `openBeneath` + close | 0.9812 | 0.9750-0.9930 |

A/A controls stayed centered at 1.00 and inside the same 5% gate. Scope is warmed steady-state Windows operation for these two bridge-heavy paths; it does not claim cold-start or tail-latency performance.


steipete:
@clawsweeper re-review

steipete:
Maintainer decision: accept the documented Windows compatibility boundary. Native descriptor operations require the host executable to expose the paired libuv conversion exports; hosts lacking either export may return ENOTSUP. Restoring raw-HANDLE or unrelated CRT namespace guessing would weaken descriptor authority and is not accepted. The qualified Node 22, Node 26, and Bun 1.4.2 behavior is the supported boundary for this change.

@clawsweeper re-review

PR #370, opened 2026-09-14
fix(path): keep generated names device-safe

steipete:
## What Problem This Solves

Fixes an issue where consumers generating portable filenames or sibling temp paths could receive a Windows reserved-device alias after truncation, sanitization, or prefix composition. On Windows, names such as `CON`, `NUL`, and related aliases can address devices instead of ordinary files.

## Why This Change Was Made

The sanitizer now rechecks device safety after its existing surrogate-safe 200-code-unit truncation, reserving one unit for the existing underscore disambiguator only when needed. Temp-file names receive the same completed-name protection, while sibling callback paths fail with `invalid-path` before hooks or producers run if their composed component is still a device alias. The checks run consistently on every platform so generated names remain portable.

## User Impact

Generated names remain within their existing length bound and valid short names are unchanged. Previously accepted caller-supplied sibling prefixes that compose to a Windows device alias now fail before producer execution instead of reaching filesystem APIs.

## Evidence

- Added boundary coverage for 199, 200, and 201 UTF-16 code units, padding, extensions, mixed case, the full reserved-device set, and surrogate pairs.
- Added real temp-workspace and ordinary/isolated/private sibling regressions proving safe output or producer-not-called rejection.
- Codex Astra xhigh exact-tree challenge: ACCEPT, no P0-P2 source findings.
- Windows Node 22 focused runtime proof: `test/windows-device-filenames.test.ts` passed 70/70, including safe temp-file creation, ordinary/isolated/private callback rejection before producers, and unchanged default publication behavior.
- Ubuntu Node 22 CI runtime proof: the same focused file passed 70/70; the complete Ubuntu coverage lane passed 9,527 tests with 133 skips.
- The complete exact-head CI matrix is green across Windows, Ubuntu, and macOS, including Node 22/24, coverage, package smoke, native checks, clippy/audit, analysis, and benchmark smoke.
- Targeted A/A plus interleaved A/B performance qualification is still pending the coordinated execution window and remains the final merge gate.


steipete:
Maintainer decision: accept the documented cross-platform rejection of callback components that are Windows device aliases. These APIs construct portable generated filenames; accepting a prefix on POSIX that cannot name an ordinary file on Windows creates a latent platform-dependent failure. The new boundary is fail-fast (invalid-path) before hooks/producers, and the migration is to choose a device-safe prefix. Runtime and performance proof remain merge gates.

PR #372, opened 2026-09-14
fix(security): bind Windows ACL checks to open files

steipete:
## Summary

- bind Windows secure-file ownership, locality, and ACL inspection to an owned duplicate of the exact retained reader handle
- fail closed when the descriptor capability is missing, stale, malformed, or unsupported; never fall back to pathname ACL inspection for secure reads
- add native and policy regressions, required-native CI coverage, and baseline-aware method-benchmark contracts

## Security

Previously, `readSecureFile` opened and pinned one object but later inspected ACLs through its pathname. A rename-capable peer could substitute a separately trusted pathname target during the ACL query, allowing the retained original bytes to be admitted using another object's permissions.

The new helper resolves the Node fd only through the host libuv bridge, duplicates its Windows handle, and obtains both the Node-compatible identity projection and security facts from that duplicate. The borrowed descriptor remains open and its file position is unchanged. Incomplete ACL layouts, remote locality, foreign ownership, and capability/version skew fail closed before bytes are read.

## Compatibility

Windows secure reads now require the matching native descriptor-inspection capability. Native-off or stale-helper configurations return `permission-unverified`; POSIX behavior is unchanged. The benchmark harness distinguishes that current contract from an explicitly selected legacy distribution's prior pathname behavior.

## Validation

- `git diff --check`
- three independent GPT-6 Astra xhigh challenge/fix cycles; final exact-diff verdict: ACCEPT, no P0-P2 findings
- `native/src/windows.rs` is byte-identical to exact base `1897be8`, preserving the host-libuv-only bridge from #368
- remote cross-platform/native CI will run on this PR
- focused local Windows runtime, handle-leak, and interleaved performance evidence remains required before merge

## Performance gate

Measure successful legacy and descriptor-bound reads separately on matching Windows native builds for 128 B through 1 MiB files and representative ACL sizes/concurrency. Investigate median regression over 10% and 50 microseconds or p95 over 20% and 100 microseconds. Report native-off rejection latency separately; it is not a successful-read speedup.


steipete:
Exact Windows method evidence from run 34912265430: the changed native-required secure-read owner improved from 670.74 us (legacy pathname ACL) to 636.18 us (descriptor ACL), -5.15%, with overlapping sample ranges. Adjacent required-mode rows were neutral or faster: readLocalFileSafely -1.08%, readRegularFile -1.42%, sync -5.01%, readOwnerAndDacl -3.71%, inspectWindowsAcl +4.36%. Native-off showed broad process/system drift (including +10.33% in unchanged inspectWindowsAcl at roughly 0.4-0.45 s) and does not exercise the descriptor-ACL change, so it is not attributed to this patch. Performance gate is provisionally clear; merge still waits for live same-handle ACL, payload/concurrency, native-off rejection, and 10k handle-growth proofs.

PR #374, opened 2026-09-14
fix(security): bind private directory creation

steipete:
## Summary

- create Windows private directories exclusively beneath a retained parent handle with the protected DACL applied in `NtCreateFile`
- validate ACL, type, locality, complete `FILE_ID_INFO`, and final public-path association against the retained created object
- clean up only through the created handle, preserving pathname replacements and combining cleanup failures

## Security

The previous implementation created by pathname, reopened the pathname for ACL inspection, and deleted by pathname on validation failure. Under mutable ancestry, those operations could describe or remove a different directory than the object created by the call.

This change retains creation ownership before any fallible query, verifies the complete 64-bit volume and 128-bit file identity, and performs nonrecursive handle-only cleanup. It preserves #368's host-libuv fd bridge unchanged.

## Compatibility

The helper still returns a pathname, so validation is point-in-time; callers must keep later pathname ancestry trusted. Immediate parent/final leaf reparses and ambiguous dot or trailing-dot/space spellings reject before creation. Earlier ancestor aliases may still resolve and are documented.

## Validation

- `git diff --check`
- independent GPT-6 Astra xhigh exact-diff review: ACCEPT, no P0-P2 findings
- focused source regressions cover parent substitution, empty and populated replacements, high identity bits, reparse/collision/pathname cases, and cleanup failures
- remote Windows native CI and method measurements will run on this PR
- same-host native performance evidence remains required before merge

## Performance gate

Compare native-required `createPrivateDirectory`, `readOwnerAndDacl`, `inspectWindowsAcl`, and `inspectPathPermissions` on the exact base and head, interleaved on the same Windows host and volume. Report medians and spread; creation intentionally adds association queries.


steipete:
Performance gate: the exact Windows method audit shows a real direct-owner regression in createPrivateDirectory: baseline median 372.92 us versus candidate 576.19 us (+54.51%), with tight non-overlapping ranges. Higher-level owners remain approximately neutral (writeSecretFileAtomic -0.12%, createSecretFileAtomic +2.93%, FileStore private writes approximately neutral), but the public primitive itself is not acceptable at this cost. Holding merge while Astra reviews a security-preserving fast path; I will rerun the exact owner benchmark after the repair.

steipete:
Security-preserving performance repair pushed as 1ad3742. Independent Astra xhigh review ACCEPTED with no P0-P2: four attribute-only queries now use FileAttributeTagInfo; five full FileIdInfo association checks and four locality validations remain; private validation avoids report-only SID strings; GUID/opened-name locality falls back to the existing normalized-DOS proof; public-parent reopen requests metadata rights only. Exact Windows 20x7 direct-owner audit is running: https://github.com/openclaw/fs-safe/actions/runs/34919346374. Merge remains held until live Windows tests and createPrivateDirectory itself show no regression.

steipete:
Performance gate remains blocked at exact head 1ad3742. The completed Windows 20x7 direct-owner audit measured createPrivateDirectory at 366.62 us baseline versus 556.14 us candidate (+51.69%); ranges do not overlap (360.13..387.89 vs 545.74..703.48 us). Standard CI is green and the prior Astra correctness review accepted the security invariants, but this public primitive is still materially slower. I am holding merge and sending the exact syscall/validation shape back through Astra for a new security-preserving redesign before changing code. Audit: https://github.com/openclaw/fs-safe/actions/runs/34919346374

PR #375, opened 2026-09-14
fix(security): validate Darwin clone ACLs

steipete:
Darwin clone admission now reads mode, ownership, exact identity, flags, and ACL state from retained descriptors. It rejects extended ACL entries on the target parent and private staging directory, verifies the normalized payload before publication and handoff, and fences the published name against replacement.

Once a clone has created bytes, normalization or security-check failures remain terminal instead of triggering an ordinary-copy retry. Unsupported admission before creation can still use the documented fallback. The internal ACL inspector preserves the caller's descriptor and file position; unsupported or malformed ACL data is never treated as absence.

Integration preserves the current Windows native capabilities and benchmark registrations and fixes a Rust pointer-type inference error. The package and native addon build successfully. Validation passed 103 Rust tests and 79 JavaScript tests (4 platform skips), including real macOS ACL and native filesystem checks. Fresh independent review through P2 is clean. The updated head will complete the platform CI matrix before landing.


steipete:
Exact-head real-runtime and performance gates are now complete.

- macOS 15.7.9 ARM native run: 60/60 native Rust tests passed; the live ACL tests confirmed populated ACL detection, pre-materialization rejection with an empty destination, descriptor retention across pathname replacement, and payload ACL removal. Required Node and Bun lanes each passed 33/33.
- APFS: the native-required audit registered and completed `copyTree/always`; Darwin's native probe returns a backend only when `fstatfs` reports APFS. The harness verified all copied names and bytes after every sample.
- Performance repeated independently on the same macOS runner class. Relevant native-required median deltas (first, repeat): `probeTreeClone` (+0.44%, +3.86%), `createCloneSource` (-36.77%, -20.77%), `copyTree/auto` (-22.97%, -25.73%), `copyTree/always` (+3.77%, -45.51%). No positive result crossed the documented 10% threshold, and the first-run `copyTree/always` drift reversed strongly.

The PR body now links the exact CI and both raw artifact runs. @clawsweeper re-review

PR #377, opened 2026-09-15
fix(security): bind removal and move parent admission

steipete:
Nonrecursive removal could resolve a path, lose its parent association, and later remove an entry through a replaced parent. This change retains exact parent receipts through resolution and rechecks them before removal, preserving alias policy and failure behavior.

The same review exposed two native-move admission gaps: a late parent alias could bypass explicit symlink rejection, and a parent already observed outside the Root could be accepted. Both selected endpoints are now admitted inside Root, explicit rejection rechecks the full operation routes, and authority callbacks are followed by fresh Root and parent fences. Default and contained-parent-following behavior remain supported.

The package builds and 145 focused removal, move, read, and metadata tests pass (1 platform skip). The move regressions use real descriptors and reproduce both gaps before the fix. Fresh independent review through P2 is clean. The updated head is undergoing complete CI before landing.


steipete:
Performance gate failed on the exact reviewed head `d2090f2707fc3d2b6444659f32126a795268e050`; this PR is held and will not merge as-is.

Repeated 7-sample method-audit medians for nonrecursive `Root.remove`:

| Platform / mode | depth 1 | depth 8 | depth 32 |
|---|---:|---:|---:|
| Ubuntu / native off | +15.25% | +120.37% | +588.49% |
| Ubuntu / native require | +1.67% | +111.64% | +606.48% |
| macOS / native off | +21.35% | +154.37% | +689.31% |
| macOS / native require | -3.40% | +152.70% | +665.26% |

The cross-platform, depth-dependent shape is a real regression rather than runner noise. A GPT-6 Astra xhigh redesign consultation is in progress to preserve the Root→parent identity fence without the apparent quadratic ancestry work. Any repair will get a fresh exact-diff Astra challenge and a new base/head audit before merge.

[Exact all-platform method audit](https://github.com/openclaw/fs-safe/actions/runs/34913844556)

steipete:
Performance repair pushed as bc87d0b. The nonrecursive path now uses one full Root guard, bigint identity-only intermediate receipts, and one full immediate-parent guard; pre/post verification keeps canonicalization constant with depth and identity work linear. Independent Astra xhigh re-review ACCEPTED with no P0-P2. Exact all-platform 20x7 audit against 1897be8 is running: https://github.com/openclaw/fs-safe/actions/runs/34918371880. Merge remains held until shallow/depth-8/depth-32 and recursive controls show no regression and exact CI is green.

steipete:
Performance hold remains after the optimized receipt head bc87d0b. Exact 20x7 method audit: Ubuntu depth-8 +42.33% off / +32.04% require and depth-32 +75.59% / +68.23%; Windows depth-8 +74.76% / +141.68% and depth-32 +99.94% / +86.28%. These owner rows have non-overlapping ranges and are blocking. Direct Ubuntu off was neutral; several macOS rows were noisy or faster, but they do not override the deep-path regressions. Artifact: https://github.com/openclaw/fs-safe/actions/runs/34918371880 . Merge remains held while a fresh Astra design review targets the remaining duplicated ancestry observations.

PR #323, opened 2026-09-13
fix(path): reject Windows filesystem namespace aliases

steipete:
Windows stream and directory-index path spellings could reach filesystem operations whose confinement checks assumed ordinary files and directories. This change rejects those aliases at raw and canonical path boundaries across Root, standalone helpers, archives, permissions, locks, queues, and native entry points.

Supported rooted and drive-relative forms retain their existing contracts, POSIX colon-bearing names remain valid, and caller-owned paths and callbacks are snapshotted once before asynchronous work. Regression coverage includes normalization, namespace roots, changing getters, callbacks, and publication boundaries.

The submitted head passed its Linux/macOS/Windows CI. Maintainer integration preserves the secret-read policy snapshot already on main, builds successfully, and passed fresh independent review through P2. Earlier final-head Windows performance clearance was not established; source review found string admission and owned snapshots rather than new filesystem loops. The rebased head `94b80d6c8a8b3acb19db75bfe15db0800953031f` passed the full CI matrix and was fast-forwarded onto main. Linux/macOS/Windows Node checks, native/musl checks, package smoke, coverage, CodeQL, and ordinary benchmarks are green.

PR #378, opened 2026-09-15
fix(security): fence Root metadata observations

steipete:
Root stat and listing operations could collect metadata or names from a replaced descendant and then verify only the Root itself. This change captures exact target and parent observations and checks them before accepting the result, including initial missing-target failures.

An optional native observation keeps identity and canonical path tied to one retained directory handle. JavaScript receipts reuse already-admitted traversal without weakening replacement checks. Integration preserves the Windows namespace/path adapters from #323 and mutation ownership from #379.

The integrated package builds and 124 focused tests pass, with 3 platform/native skips. Fresh independent review through P2 found no actionable issue. The updated head `5108b3bb6250631357485b1ff068cedc0fb910f9` passed the complete CI matrix and the Linux/macOS/Windows public mutation proof, then was fast-forwarded onto main.


steipete:
Performance gate failed on exact head 5f13e84 (50x7 method audit 34915005020). Direct owners show structural regressions: Ubuntu Root.stat/nested +71.72%/+64.98% (off/require), depth-8 +70.39%/+68.57%; one-entry list +23% to +41%. Windows stat/nested +225.76%/+214.14%, depth-8 +112.71%/+115.94%; one-entry list about +97% to +103%, depth-8 +66% to +73%. Most ranges are non-overlapping; large-list metadata rows amortize the fixed cost but do not clear the primitive. Merge is held while Astra redesigns the containment receipt/path-owner flow. No weakening will be accepted to buy back latency.

PR #379, opened 2026-09-15
fix(security): bind pinned writes to mutation policy

steipete:
Pinned writes could resolve a different parent after their initial policy check and reuse stale authorization. This change binds snapshotted mutation policy to the actual selected parent and destination, retaining exact Root and directory identities through component creation, staging, and publication.

The same admission applies to shared JavaScript writable opens, append, mkdir, and Windows buffered writes. Operation-local receipts avoid repeating complete traversal where exact identities and unchanged routes prove the existing admission still applies. POSIX native code supplies direct-child creation provenance; TypeScript continues to own policy. The documented Windows-native recursive-parent limitation remains unchanged.

Maintainer integration preserves the Windows namespace checks and single-read snapshots from #323. Mutation and namespace regressions passed together (120 tests, 18 platform/native skips), the package builds, and fresh independent review through P2 found no actionable issue. Internal write types were extracted to keep the implementation within its existing size budget. The rebased head `2a521d5c131ed5d89fdb197994e4b09ad5d4adb7` passed the complete CI matrix and the dedicated Linux/macOS/Windows public mutation proof, then was fast-forwarded onto main.


steipete:
Performance gate failed on exact head 65caf65 (run 34916227621; configured 20 iterations/7 samples, dedicated rows use divisor 10 = 2 operations/sample). Ubuntu native-off existing-parent writes regress +35.9%/+127.4%/+323.5% at depths 1/8/32; missing-parent writes +125.7%/+489.8%/+898.8%. Native-required missing-parent rows are +138.2%/+830.5%/+3115.6%. macOS is similarly structural. Windows off improved 12-17%, but native-required remains +31-47%. Ranges are predominantly non-overlapping. CI success does not clear this. Merge is held while Astra redesigns the policy-admission receipt; every actual mutation and policy revalidation must remain bound.

PR #380, opened 2026-09-15
fix(security): bind Windows Root prefixes by identity

steipete:
## Summary

- reject case-fold-only Windows Root prefixes unless the supplied prefix has the retained Root identity
- rebase admitted paths to trusted Root spelling across reads, writes, publication checks, listings, mkdir, and both move endpoints
- preserve ordinary resolver receipts, redundant-separator compatibility, and final-symlink rejection
- add modeled zero-observation tests, real case-sensitive-directory coverage, docs, changelog, and paired benchmark rows

## Security

Windows path comparison is normally case-insensitive, but a case-sensitive directory can contain Root and root as distinct objects. Lexical folding previously admitted the sibling. Exact structural prefixes remain allocation-only; only a match that depends on folding performs bounded identity admission.

## Performance

Source-confirmed exact-prefix overhead: zero filesystem/native/subprocess/probe calls. Identity admission with the retained bigint Root receipt performs one candidate observation normally and at most two for an ambiguous identity. The final I/O admission is exact after rebasing and adds no observation. Real Windows end-to-end measurements are required before merge.

## Review and gates

Codex Astra xhigh challenged the implementation repeatedly and found four compatibility/build issues plus one normalized-source confinement escape. All were fixed; its final full-diff review ACCEPTED with no P0-P2. Local runtime commands were intentionally not run while shared test resources were reserved. Exact CI, non-skipped case-sensitive Windows proof, and interleaved performance comparisons remain merge gates.


steipete:
CI-blocker repair pushed as 1e0ee60. The move preflight was extracted verbatim into a lower-level internal module (root-impl.ts now 1749/1750 lines), and only the three POSIX hybrid fixtures that spoof win32 temporarily restore the captured host platform while invoking the real synchronous Root admission. Independent GPT-6 Astra xhigh exact-diff review: ACCEPT, no P0-P2; denial order/options and call position are unchanged, the adapters capture the original function before spying, restore the full descriptor in finally, and preserve all existing error/retry/closure/preservation assertions. No runtime was run locally under the serialized resource hold; exact CI is now the execution gate. The production extraction adds no calls, awaits, promises, or filesystem work.

PR #382, opened 2026-09-15
fix(security): sanitize sibling callback names

steipete:
Fallback filenames pass through the same sanitization as primary names, and completed sibling staging components are validated before producer callbacks run. This closes bypasses through fallback and callback names while preserving bounded filenames, extension handling, and Windows device-name rejection.

The integration keeps the current descriptor-handoff and private-producer cleanup lifecycle. Callback parents retain valid Windows namespace-root spelling, and raw/resolved parent aliases are checked before constructing the completed path. Internal callback tests use the new directory/component interface while preserving receiver identity and isolation behavior.

The package builds and 218 focused filename, output, callback, snapshot, namespace, and private-producer tests pass (26 platform skips). Fresh independent review through P2 is clean. Complete platform CI passed on the final PR head.


steipete:
Maintainer decision: accept the hardened exported fallback contract, conditional on the promised consumer proof and performance evidence. A fallback that is already a safe bounded filename remains its existing fixed point. Empty, reserved, separator-bearing, colon-bearing, trailing-dot/space, control-bearing, or overlong fallbacks may now be sanitized, truncated, or replaced with the documented fixed ile default instead of being returned verbatim. That compatibility change is intentional: fallback text is attacker-controlled at the same boundary as the primary name and must not bypass the shared filename policy. Callers that require an exact display label should store it separately from the filesystem component. Landing remains held until exact-head public-API evidence covers safe-caller compatibility, confined ordinary and private publication, and rejection before producers, and until the comparative performance gates pass.

PR #384, opened 2026-09-15
improve(copy): reuse directory admission metadata

steipete:
## What Problem This Solves

Portable directory copying repeats a child directory lstat and allocates unused Dirent objects. Carry the classification stat into recursive admission and enumerate names while retaining the opened-handle identity checks, per-entry classification, cancellation, settlement, and metadata fences.

## Why This Change Was Made

Remove repeated work inside the existing owner without changing the public API or filesystem identity checks.

## User Impact

Not selected for landing: fewer source-level observations did not produce a reproducible performance benefit.

The Windows Node 26.8.2 / NTFS matrix covered flat256 and nested64 fixtures, auto/never policies, workers 1/4/16, four ABBA blocks per cell (240 total copies including warmups). Every copy passed bytes, inventory, directory timestamps, and destination-mutation/source-independence checks outside timing. Actual matching addon hashes and distinct JS/WASM builds were verified; no samples excluded; owned fixtures were cleaned.

Results were mixed, with large outliers. The only cell positive in every initial block was nested64/auto/16. A separate fresh-process confirmation deliberately retested that selected cell with five ABBA blocks: baseline median 976.37 ms versus candidate 1248.08 ms, with four of five blocks slower and paired median gain -17.48%. All 24 confirmation copies and cleanup checks passed. The initial apparent benefit did not reproduce.

A [Linux native-off mixed-copy control](https://github.com/openclaw/fs-safe/actions/runs/34934565332), Node 24.20.0, used the same harness and three ABBA blocks. Paired gains were -3.91%, -11.08%, and +3.81%, likewise not consistent. Its workload differs from the Windows directory-heavy case.

## Evidence

31 focused copy/clone tests passed, 14 platform skips. Fresh independent review: no P2-or-higher findings.

Build, static checks, docs, and separate package check passed. Full local check: 8,180 passed, 69 failed, 1,346 skipped, one fixture error. Exact failure headers are a subset of unchanged main at `1fca788` (8,179 passed, 70 failed, same skips/error). All 15 [cross-platform CI jobs](https://github.com/openclaw/fs-safe/actions/runs/34933956573) passed. Correctness is supported, but the performance evidence does not justify landing this candidate.

- [x] Tests reviewed for observable contract coverage
- [x] `CHANGELOG.md` updated
- [x] Outgoing content checked for private data


steipete:
Deferring this optimization after measurement: the initial positive nested-copy cell did not reproduce in a fresh process, and the Linux control did not show a consistent benefit. Correctness and all cross-platform CI jobs passed. The body records both the original experiment and the negative confirmation; this change is not in the performance landing set.

PR #386, opened 2026-09-15
improve(windows): reuse directory enumeration storage

steipete:
## What Problem This Solves

Windows native tree traversal allocates and zeros a 64 KiB enumeration buffer for each directory. Reuse one operation-owned aligned buffer in ReFS copying and bounded temp-workspace cleanup. Returned names own their storage, and every directory read restarts enumeration.

## Why This Change Was Made

Remove repeated work inside the existing owner without changing the public API or filesystem identity checks.

## User Impact

Not selected for landing: source-level allocation churn is lower, but the measured public cleanup path did not show a repeatable latency improvement. ReFS is unavailable on the local host, so there is no measured ReFS benefit to justify this change yet.

Windows Node 26.8.2 / NTFS, public TempWorkspace.cleanup with require-bounded, 2,001 directories and one file per workspace, five ABBA blocks:

- Baseline median 499.41 ms (389.63–864.50); candidate 500.73 ms (362.60–947.65), effectively unchanged.
- Candidate block means were faster in two of five blocks and slower in three. No samples excluded.
- Actual baseline/candidate addon hashes were pinned; built JavaScript/WASM manifests were identical. Fixture preparation and verification were outside timing.
- Every timed operation, warmup, and preflight passed removal, repeated-cleanup, and sibling-preservation checks; task-created fixtures were cleaned.
- A separate three-workspace proof reached exactly three native calls, removed all directories, and returned handles from a peak of 261 to the starting 257 after each cleanup.

The allocation count (2,001 scratch buffers versus one per operation) is derived from source, not allocator tracing. This evidence does not establish lower peak memory or end-to-end latency.

## Evidence

93 Rust tests passed, including reuse across large, small, and empty directories. Native/package builds and pack check passed. Fresh independent review: no P2-or-higher findings. ReFS is not mounted on the current host; native ReFS end-to-end proof is pending/unavailable locally.

Full local check: 8,180 passed, 69 failed, 1,346 skipped, one fixture error; exact failure headers are a subset of unchanged main at `1fca788` (8,179 passed, 70 failed, same skips/error). All 15 [cross-platform CI jobs](https://github.com/openclaw/fs-safe/actions/runs/34933962867) passed. Build, static/docs checks, separate package check, and all 93 Rust tests passed.

- [x] Tests reviewed for observable contract coverage
- [x] `CHANGELOG.md` updated
- [x] Outgoing content checked for private data


steipete:
Deferring this candidate: correctness and cross-platform CI passed, but paired NTFS cleanup measurements showed no repeatable latency improvement. The body records the full measurements and allocation-proof limits. No measured ReFS benefit is available on this host, so this is not included in the performance landing set.

PR #352, opened 2026-09-14
feat(path): probe aliases of missing path suffixes

steipete:
## What Problem This Solves

Applications comparing paths before files exist need to observe case and Unicode alias behavior in the selected directory. OpenClaw's agent database registry currently owns temporary-name generation, nested probes, filesystem identity checks, and cleanup alongside its actual ownership policy.

## Why This Change Was Made

Add `probePathSuffixAliasesSync()` and `ProbePathSuffixAliasesOptions` under `@openclaw/fs-safe/advanced`. The helper probes complete missing suffixes using owned temporary directories and returns `true`, `false`, or `undefined`. The trusted synchronous predicate lets callers retain their own case-variant eligibility policy, evaluated in component order. The library keeps no cache.

Generated names avoid normalized case aliases of the requested components. Exact raw Unicode pairs are confined to owned neutral directories. Root and ancestor identities are checked before mutations and after observations; cleanup only removes verified empty directories. Any failed observation or cleanup makes the result unknown, including transient root disappearance. Callback errors propagate after cleanup.

This requires a writable observation directory. It preserves existing parent permissions and uses ordinary mkdir mode/umask. Creation and first observation, and each final check/syscall, remain separate; an unverified or replaced probe may remain. This does not reserve paths or authorize later access.

## User Impact

The prepared OpenClaw companion removes **308 production lines standalone**, or **309 additional lines after #349**. Together the two proposals remove **429 production lines**, with **zero test deletions**. The registry retains its cache, ownership decisions, dotted-I exclusion, and unknown-as-collision policy.

The companion is bound to OpenClaw `0cede6c5046c858b6e7de4fe6f30bb180fb15d1f`; its three affected files are byte-identical and the combined patch still applies at `2545695c28b5a036ab678c4fed377c47c8adfb0a`. Adoption waits for a package release containing these APIs. No unavailable imports or dependency overrides are committed to OpenClaw, and this proposal includes no release.

fs-safe adds 351 production lines for this primitive and shared-helper cleanup. Including #349, total production code across both repositories grows by seven lines; the 429-line reduction is specifically in OpenClaw.

The stricter unknown result after failed cleanup is deliberate. Performance is workload-dependent: early policy exclusion avoids filesystem setup and unnecessary randomness, while real nested probes retain identity-check costs. The registry's cached path is unchanged.

## Evidence

[...]

PR #414, opened 2026-09-16
feat(path): add bounded suffix alias probes

steipete:
Add opt-in bounded probes for aliases of missing path suffixes. Fixed path, depth, creation, and observation limits bound the work; exact identity checks and conservative cleanup preserve entries the operation cannot prove it owns. Documentation makes the remaining pathname race limits explicit. This supersedes #352.

The observation preserves identical-suffix no-I/O behavior, raw Unicode comparisons, caller policy, and exact thrown callback values. Replaced or nonempty owned directories are preserved during cleanup. Valid Windows namespace roots retain their separator and use the filesystem dispatch adapter without changing the caller-owned admission contract.

The package builds, 135 focused suffix/bounds/prefix/case tests pass (22 platform skips), and docs validation passes. Installed-package smoke passes with npm and pnpm on macOS, including both native settings, collision exhaustion, resource limits, callback replacement, and cleanup cases. Fresh independent review through P2 is clean. Complete platform CI passed on the final PR head.

An intermittent native macOS contention failure passed its unchanged-source retry. Investigation then reproduced a separate two-release metadata race; the guarded fix and regression coverage landed in [#430](https://github.com/openclaw/fs-safe/pull/430).


steipete:
## Performance HOLD — exact production/benchmark blobs at `d03b270`

The four planned pilot campaigns completed and were independently audited from raw artifacts:

- source ABBA: https://github.com/openclaw/fs-safe/actions/runs/35042424582
- source BAAB: https://github.com/openclaw/fs-safe/actions/runs/35042424637
- same-artifact control ABBA: https://github.com/openclaw/fs-safe/actions/runs/35042424870
- same-artifact control BAAB: https://github.com/openclaw/fs-safe/actions/runs/35042424724

Provenance passed: 12 studies, 288 reports, 504 measured rows, 72 expected skips, 3,528 samples; all plan/report/runner-output hashes, source bytes, before/after installations, OS/mode/order positions, and addon identities matched. Candidate/harness tree was `9fb320919ee2652a035e871d37a4b07ca3a70dc1`; baseline tree was `d7e80b18ac236471e50a45a045d53d3c4989cc0c`.

Strict result: **BLOCK**. The unchanged shared case-probe comparison breached 4/48 source cells (one pooled, three block); controls independently breached 6/96 cells and were not subtracted. Source blockers included macOS/off ABBA block-3 median +14.68%, Windows/off ABBA pooled/block-3 max +56.70% (+309.10 µs), and macOS/require BAAB block-1 median +18.39% / max +26.23%.

Candidate-only suffix pooled medians ranged from 52.8–164.6 µs on Linux, 196.6–238.5 µs on macOS, and 1.18–1.38 ms on Windows depending on order/mode. These are shallow ASCII-case observations, not general Unicode/depth/startup claims.

Commit `12bdbc8` adds only packed-consumer proof files and CI wiring; production and benchmark blobs are unchanged. I will not selectively rerun an unchanged measured implementation after a failed gate. This PR remains open but must not merge under the established performance policy.

steipete:
Independent GPT-6 Astra xhigh audit of the exact installed-package artifacts: ACCEPT for 12bdbc8ba6b49a797f644462eccd9d682943dc84 / tree 45e4ac9433f5270bc8cedbd9844cd71a56413a3a.

Verified all six artifact ZIP digests and all 12 JSON entries; exactly 24 npm/pnpm × native off/require receipts and 288 ordered scenario rows. Source/head/tree/clean projections, tarball integrity chain, 21 compiled-module hashes, LF/CRLF probe/helper hashes, addon-mode evidence, callback/collision/ancestor-swap/nonempty cleanup witnesses, and scenario ordering are internally consistent. Native priming is honestly separated from suffix execution: suffixNativeExecutionClaimed is false.

Artifacts retain JSON receipts rather than tarballs/binaries, so those bytes are bound by CI's reviewed hash chain but cannot be rehashed from the artifact alone. finalInventoryEmpty follows deliberate harness fixture cleanup; per-row cleanup receipts retain the preservation evidence.

Exact proof jobs all passed. Node 24 current vs immediate-parent job/proof-step durations were Ubuntu 70s/11s vs 86s/13s, macOS 69s/11s vs 81s/14s, Windows 164s/26s vs 173s/44s. This is descriptive CI timing, not a controlled benchmark. Expanding to Node 22 doubled aggregate proof capacity (652s across six jobs vs 340s across three).

Preserved audit directory: C:\Users\Peter Steinberger\AppData\Local\Temp\fs-safe-pr414-astra-proof-185ca82a2ca74f84af47535baab06447

The raw package-proof gap is closed. The separately documented strict source-performance BLOCK remains binding, so this PR stays on hold and should not merge.

PR #415, opened 2026-09-16
fix(security): stop compressed reads after cancellation

steipete:
Native zstd and bzip2 readers check cancellation before refilling compressed input. Existing decoded-output checks remain in place, so cancellation also stops reads hidden inside a decoder without weakening framing, concatenated-stream, or complete-archive validation.

Regression coverage aborts during the first raw read for both file and memory sources, verifies that no second refill occurs, checks cancellation with buffered decoded output, and retains valid/corrupt/truncated concatenated-stream checks. The native benchmark fixtures cover the affected readers.

The package and native addon build successfully. All 109 Rust tests and 87 focused cancellation, native TAR, compressed-filter, framing, and PAX tests pass. Fresh independent review through P2 is clean. Complete platform CI passed on the final PR head.


steipete:
Performance gate: BLOCKED. Independent raw-artifact recomputation finds 234/384 source failures and 211/384 same-artifact control failures. Correctness and mutation evidence remain accepted, but the no-slowdown requirement is not proven; this PR stays unmergeable.

PR #417, opened 2026-09-16
fix(security): retain exact copied directory identities

steipete:
Copy-fallback cleanup could remove a directory replaced while child cleanup yielded. A numeric copy manifest also allowed a different inode that rounded to the same number to be admitted as the cleanup target.

The move fallback now retains a private exact directory receipt from copy admission through traversal, publication, and cleanup. Cleanup compares against that receipt before removing children and rechecks it after the mutation-authority callback immediately before directory removal. Observed replacements and persistent unknown identities fail closed; stale source changes are reported without deleting the replacement.

The directory receipt stays separate from leaf timestamp and hardlink-cleanup accounting. Documentation covers the exact admission and recheck points and the remaining pathname syscall gap.

The package builds and all 114 focused move, directory-identity, authority, and hardlink tests pass (7 platform skips). Regressions cover rounded-equal replacement before cleanup, replacement by the final authority callback, and unknown identity before copying. Fresh independent review through P2 is clean. Complete platform CI runs before landing.


steipete:
Security validation update for exact head 8704e6770c3dd9c48d49c82f6e4a0781974d2851 (tree 28939c976cde5941a7f35195c255462d3d6273cf):

- Full PR CI is green across Node 22/24, Linux, macOS, Windows, native, musl, coverage, package smoke, CodeQL, and workflow lint.
- Positive baseline-sensitive run https://github.com/openclaw/fs-safe/actions/runs/35047323887 passes all relevant lanes.
- Production-only mutant cd8504473f610c51e5556db442f605f697e1ffc1 (tree ee1437bbdd64f02b4c99ac66d68e564d549b1a57) restores only src/move-path-cleanup.ts to its base blob. Run https://github.com/openclaw/fs-safe/actions/runs/35047409246 then fails exactly the two targeted suites on every Node 22/24 Linux/macOS/Windows JS lane: 8/14 cleanup cases and 15/15 identity cases. A fresh independent GPT-6 Astra xhigh audit accepted the proof with no P0-P2 findings.
- Four-order performance runs:
  - source ABBA https://github.com/openclaw/fs-safe/actions/runs/35047101789
  - source BAAB https://github.com/openclaw/fs-safe/actions/runs/35047104433
  - same-artifact control ABBA https://github.com/openclaw/fs-safe/actions/runs/35047106612
  - same-artifact control BAAB https://github.com/openclaw/fs-safe/actions/runs/35047108350

The independently recomputed artifact audit verified all 16 ZIP digests, 12 studies, 288 reports, 864 workload rows, 6,048 samples, and 288 per-block/pooled gates with zero provenance or calculation errors. Source comparisons fail 84 gates, while the controls independently fail 79 gates (31 ABBA, 48 BAAB). Because controls must pass and cannot be subtracted, this is inconclusive rather than a source-attributable regression.

Merge status: HOLD. The strict no-slowdown acceptance criterion is not established, so this PR will not be landed from this evidence.

PR #418, opened 2026-09-16
fix(security): pin secure temp directory repairs

steipete:
Secure temporary-directory permission repair now opens a no-follow directory descriptor and binds every repair to exact device, inode, owner, and type receipts. It rechecks the descriptor and named directory around chmod and access checks, closes the descriptor on failures, and preserves concurrent replacements. A failed chmod can be accepted only when fresh observations prove that a concurrent repair made the original directory safe.

The resolver snapshots its hooks and policy at entry. A complete descriptor adapter is required for injected repair authority; legacy pathname hooks cannot silently authorize host chmod or creation. Windows keeps its documented no-chmod behavior and namespace admission. The exported adapter type, docs, and package API receipt are updated together.

The package builds and 157 focused tests pass (3 platform skips), including real permission repairs, restrictive umasks, descriptor cleanup, replacement races, and Windows namespace boundaries. Package validation, including the installed guest protocol, passes. Fresh independent review through P2 is clean. Complete platform CI passed on the landed head.


steipete:
Performance gate: BLOCKED at the current head. Independent GPT-6 Astra xhigh review reproduced 139 source failures and 37 same-artifact control failures under the stated OR thresholds. I am retaining the correctness proof but will not merge this implementation; the PR body now records exact counts, skipped Windows coverage, and the redesign requirement.

PR #420, opened 2026-09-16
fix: close sync lock descriptors after stat failures

steipete:
Synchronous lock cleanup now closes its descriptor even when metadata inspection fails. It preserves the original inspection error and leaves an unverified sidecar untouched, avoiding both a descriptor leak and unsafe pathname deletion.

Descriptor ownership is consumed before close, preventing a failed close from causing a second close of a reused descriptor number. When closing or verified removal also fails, the combined error retains the original acquisition failure.

The package builds and all 52 focused acquisition-cleanup, reentrancy, namespace, and Windows-denial tests pass (5 platform skips). Fresh independent review through P2 is clean. Complete platform CI runs before landing.


steipete:
@clawsweeper re-review

The PR body now contains exact candidate and base-production-mutant proof. Independent Astra review found all nine exercised Node/Bun mutant lanes fail exactly the five focused cases with no incidental failures. Performance remains pending, so the PR is still held.

steipete:
HOLD on exact head `2594f223dc1a8380c00cb6e0c1b6dd6e1f6c3866`: correctness, cross-platform CI, and unchanged-production kill proof are accepted, but performance certification is blocked.

Fresh GPT-6 Astra xhigh independently validated all raw evidence and recomputed **52/96 failed gates**: source ABBA 9, source BAAB 13, same-artifact ABBA 14, same-artifact BAAB 16. The same-artifact failures make attribution inconclusive, and the benchmark covers successful uncontended acquisition rather than the changed failure-cleanup path. Static normal-path overhead should be neutral; the measurements still do not satisfy the required policy.

Evidence: `C:\Users\Peter Steinberger\AppData\Local\Temp\fs-safe-pr420-independent-audit-6954a4ccd93d48cfac219060d5d7edd2`. No failed control/block/order is being subtracted or discarded. Do not merge this revision.

steipete:
A fresh GPT-6 Astra xhigh performance-design consultation confirms this exact head must remain held under the fixed gate.

Static comparison found no added successful-path source operation: only the import and fd-owned failure cleanup changed, while the existing lifecycle benchmark never enters that catch path. That supports a narrow source-relevance statement, not a latency certificate—the enclosing function body and emitted module still differ, so compilation effects remain possible. It cannot retroactively exclude the 52/96 failed source/control gates.

A deterministic failure-region benchmark could add useful diagnostics, but the candidate necessarily performs a real close that the vulnerable baseline omitted and may construct the required suppressed error. Those security-required failure operations cannot honestly be optimized away or guaranteed to fit the relative micro-gates. New measurements also cannot erase the historical fixed-gate failure.

**Decision remains HOLD / do not merge.** No defensible production change was identified that preserves close-once/fail-closed behavior and certifies every existing gate.

PR #426, opened 2026-09-16
fix(secret): snapshot mutable read policies

steipete:
Secret reads could observe caller changes to ownership, permission, link, or filesystem-adapter options after asynchronous I/O began. This change snapshots those policies at entry and uses the same admitted policy through the complete read, preserving existing error and link behavior.

The fix leaves filesystem operation counts unchanged. Regression tests exercise mutation during suspension; fresh independent review through P2 found no actionable issue. Focused local tests passed (8 tests), and the exact PR head passed Node 22/24, Linux/macOS/Windows, native/musl, package, coverage, and CodeQL checks. Earlier strict benchmark failures also occurred on identical-artifact controls and do not establish a causal regression in this change.


steipete:
HOLD: exact-head correctness and mutation proof are green, but strict performance acceptance is red.

Independent Codex Astra recomputation of the four balanced studies found **327 pass / 57 fail across 384 independent gates**:

- source ABBA: 91 pass / 5 fail
- source BAAB: 80 pass / 16 fail
- same-artifact control ABBA: 80 pass / 16 fail
- same-artifact control BAAB: 76 pass / 20 fail

That is 21 source failures and 36 control failures. All 288 reports, 576 workload results, and 4,032 raw samples were present; report and provenance seals matched. The controls demonstrate noise but are not subtracted or used to excuse a source failure. No selective rerun, exclusion, cross-order pooling, or report-median averaging is used.

Worst source maximum: macOS/require `readSecretFile`, BAAB block 3 and pooled, 226.697220 → 926.118130 µs (+699.420910 µs / +308.526461%).

Do not merge this head under the strict performance policy.

PR #427, opened 2026-09-16
fix(temp): add bounded temp-file cleanup

steipete:
`tempFile()` and `withTempFile()` gain opt-in `cleanupSafety: "require-bounded"`. The bounded path admits an existing trusted root before creation, retains the created directory identity, and shares one descriptor-based cleanup owner across manual cleanup, disposal, scoped use, and process exit. Unsupported cleanup capabilities fail before creating a child; bounded cleanup never falls back to pathname-recursive deletion.

The default remains `compatible`, including its existing error callback and cleanup behavior. Documentation explains the remaining check-to-recursive-removal race in that default and in output helpers that do not expose the new option. Integration uses the current temp-workspace admission and retained-child lifecycle while preserving exact bigint receipts and Windows namespace checks.

A construction-time descriptor leak found during integration is also fixed: initial filename validation now runs before acquiring retained resources. Four regressions reproduce rejected factories retaining resources before the fix and verify that invalid names cause no descriptor opens, directory creation, registration, or scoped callback afterward.

The package builds and package validation passes. The final focused run passed 176 cleanup, admission, and namespace tests (15 platform skips), following the broader 247-test integration run. Fresh independent review through P2 is clean. Complete platform CI passed on the landed head.


steipete:
Final evidence: **HOLD — do not merge**.

Astra accepted correctness and the two exact mutation proofs across every Node 22/24 fallback job, with no unrelated failures. The descriptor mutation proves the retained **child** fd leak through `tempFile`; the outer catch still closes the parent.

Strict raw-sample performance fails: **89/768 source gates** and **102/768 same-artifact control gates**. Controls demonstrate variability but cannot waive any source failure. All 40 ZIPs and 624 entries, 576 reports, 2,304 results, and 16,128 samples verified without discrepancy. The PR body now contains run links and the complete decision basis.

PR #433, opened 2026-09-16
fix(security): enforce native Windows parent policy

steipete:
Enforce Windows native mutation policy before each missing parent is created. Retain the originating native descriptor closer and exact parent/Root identity checks through success, failure, and Worker cleanup.

Reuse operation-local creation receipts for eligible no-follow paths, including ordinary portable slash inputs. Existing parents retain the one-open path. Opaque callbacks, aliases, incomplete identity observations, creation collisions, and unsupported optional observations keep full admission. Copy operations omit the assertion wrapper only when neither a signal nor an assertion exists.

Validation: complete Linux pnpm check; 90 focused Windows tests passed (9 platform/fixture skips), including required-native Worker ownership, parent-policy, shared-session, native-observation and wide-identity cases. Independent Codex review found no actionable P0–P2 issues.

Fresh Node 24.21.0 Windows measurements use one frozen portable-path harness and alternating ABBA/BAAB order. A policy write creating 32 parents improved from 208.9 ms in the original proposal to 37.9 ms after optimization (82% reduction); depth 8 improved from 31.0 to 13.1 ms. Against current main, the final portable-path measurements were about 37.7 versus 11.1 ms for creating 32 parents, and 7.4 versus 9.0 ms when those parents already exist. The remaining first-creation cost enforces policy checks absent from main. Deny-only or otherwise ineligible routes retain full admission. Historical rejected studies are not reclassified as accepted.

Independent code-owner approval is required by the main ruleset before merge.


steipete:
Performance evidence verdict: REJECT; this PR remains draft and must not merge.

Exact candidate: 1a3a729f8417c4864c812ce64180bba98a235289 (tree ac9537744621a2494e0c6d465b706718cbacef97), base b1987ffbe7211f9a58662069c93887f9c568b169. Candidate CI 35131714899 passed on attempt 2; coverage 35131714810 and public mutation proof 35131714729 are green. Exact mutant df06ad98f972fc30fd67161f46d09947e40bf828 makes only the two production files equal base while retaining the candidate tests; every six Node/platform fallback cell fails exactly 17 + 12 intended assertions, with no unrelated failing suite.

Astra independently downloaded and verified every byte of all 16 performance archives, 192 reports, 3,456 measured records, and 24,192 raw samples. Plans, reconstructed raw receipts, source/tree/blob identities, harness, installations, native loads, ABBA/BAAB ordering, and same-artifact controls all match. No exclusions, subtraction, or reruns were used.

Fixed-gate failures:
- sources: 401/576 rows (282 median, 301 maximum); 298/432 block and 103/144 pooled
- controls: 385/576 rows (269 median, 289 maximum); 287/432 block and 98/144 pooled
- every one of the four same-artifact control studies fails

Worst source median absolute: +35,000.125 us. Worst source maximum absolute: +286,404.400 us. Worst control maximum: +927,896.350 us. Hosted variability is substantial, but controls cannot cancel source failures and independently prevent acceptance.

Performance runs: 35131984484, 35131984455, 35131984718, 35131985340, 35131984782, 35131984773, 35131984803, 35131985388.

PR #443, opened 2026-09-17
fix(root): reduce filesystem admission overhead

steipete:
## What Problem This Solves

Fixes repeated filesystem walks and metadata queries that made guarded writes, directory-mode repair, and missing temporary-root creation unnecessarily slow after boundary hardening.

## User Impact

Consumers get faster filesystem admission with the existing confinement, identity, symlink, callback, locking, and publication guarantees. Public APIs and defaults stay the same.

## Why This Change Was Made

Windows compatibility writes now distinguish the library's live lock-destination check from arbitrary caller callbacks, allowing eligible parent admission to be reused. The historical slash-based lock key stays intact while the writer receives native path separators. Missing-path hints retain a search position and refresh their evidence; an already-complete parent keeps the cheaper ordinary leaf check.

POSIX policy writes share identical observations within a synchronous phase and can obtain fresh directory facts from retained descriptors. Windows metadata inspection avoids duplicate queries on the same handle. Store mode repair and missing temp-root creation reuse canonical results where the required final identity and canonical fences already establish them. Unknown, changed, unavailable, or malformed observations retain the appropriate rejection or guarded fallback.

## Evidence

Balanced ABBA/BAAB studies use the fixed method harness at `655be0d3145257d5e529a06998d2d23375cf1632`, explicit separate distributions, Node 24.21.0, and hardened control `6404191fd6e73bf34bcfacaefe2f113a2b8f6d99`. Samples are average time per call, with fixture setup and verification outside timing.

Representative isolated-runner results:

| Workload | Hardened control | Optimized | Reduction |
|---|---:|---:|---:|
| Linux native policy write, 32 missing levels | 15.79 ms | 9.27 ms | 41% |
| Linux sync-store permission repair, depth 16 | 3.17 ms | 1.89 ms | 41% |
| Linux sync-store new directories, depth 16 | 3.45 ms | 2.14 ms | 38% |
| Linux missing temp root, depth 32, async / sync | 5.08 / 4.33 ms | 4.11 / 3.39 ms | 19% / 22% |
| Windows policy exclusive write, 32 existing levels | 100.90 ms | 15.11 ms | 85% |
| Windows policy exclusive write, 32 missing levels | 282.70 ms | 69.22 ms | 76% |
| Windows policy overwrite, 32 missing levels | 296.70 ms | 81.24 ms | 73% |
| Windows nested native stat | 384.5 µs | 250.0 µs | 35% |

Linux final `pnpm check` passes 11,669 tests, including package checks; `pnpm test:security` passes 227 tests and `pnpm native:test` passes 104 tests. Exact-source Windows proof passes 176 targeted plus 204 security tests on Node 24, and 272 targeted tests on Node 22.23.2, including native worker ownership. Both runtimes pass package/held-lock qualification as applicable. The unchanged real ACL-denial test also passes on both runtimes in separate child tokens without Backup/Restore privileges: direct and verifier opens return `EPERM`, publication is preserved, and the original DACL is restored; parent privileges remain unchanged. The repaired namespace oracle passes its 4,000-case Windows property and explicit regression on both runtimes.

Final head `8e62f31dd67043da0d6ecd9a4c9ae97965e0e995` has green full GitHub CI across Linux, macOS and Windows, Node 22/24, native, package, coverage and public mutation-proof lanes. Its last commit is the tested namespace fixture repair; runtime production files remain those of `9b6e191`. Independent review through P2 has no accepted findings. Small local macOS timing differences are inconclusive because identical-artifact controls showed substantial host variability.

### Completed real Windows compatibility-write publication proof

The following copied terminal result was produced on **both Node 24.21.0 and Node 22.23.2**, using the final built package on real Windows. Production source files match runtime commit `9b6e191`; the subsequent commit changes only a test oracle. The JS/WASM distribution SHA-256 is `194b1448e42873deb24b9ea52b77250201f780dc56887812f6e7f5a24063980c`, verified against the downloaded build and benchmark reports.

The driver calls the public `Root.write` with `renameIdentity: "verify-content-with-lock"`, enabled mutation policy, and **no user mutation callback**. It first holds the real historical slash-key sidecar, confirms the writer has reached that lock and created no parents, releases it, waits for publication, then reads the final file back. Each `PASS` below is printed only after exact readback equals the supplied `"next"` payload and staging/lock cleanup checks pass. The last two cases additionally verify preceding-holder replacement and original-alias retargeting; the unrelated aliased file stays `"untouched"`.

```text
PASS missing-policy-string-native
PASS missing-policy-string-slash
PASS missing-policy-buffer-native
PASS missing-policy-buffer-slash
PASS refresh-target-after-previous-holder-replacement
PASS retain-selected-target-after-original-alias-retarget
WINDOWS_OPTIMIZED_COMPATIBILITY_LOCK_PROOF_PASS 6/6
```

For the four missing-parent cases, the driver verifies this after awaiting the actual write:

```js
assert.equal(await fs.readFile(path.join(directory, "parent", "deep", "value"), "utf8"), "next");
assert.deepEqual(await fs.readdir(path.join(directory, "parent", "deep")), ["value"]);
```

The balanced depth-32 Windows timing rows above separately perform a fresh content assertion after every timed write; failed publication cannot be reported as a fast sample.


steipete:
Completed real Windows publication proof is now in the PR body, bound to the final runtime source and distribution hash. On both Node 22.23.2 and 24.21.0, public Root.write waits on the actual compatibility sidecar, publishes after release, and passes exact file-content readback and cleanup checks for strings, Buffers, native separators, and slash separators. Preceding-holder replacement and original-alias retargeting also pass. Final-head CI is green across all supported platforms.

@clawsweeper re-review

PR #456, opened 2026-09-17
fix(release): require npm publish approval

steipete:
## Summary

- bind the npm OIDC publish job to the protected `npm-publish` environment
- document its independent reviewer and protected-tag contract
- add a focused structural regression test for the job-level authority boundary

## Security review

Codex Astra found that the existing environment had never been engaged: the `v0.14.0` publish succeeded through an environment-free npm trusted publisher. Astra ACCEPTS this exact repository patch, but full P2 closure requires coordinated npm configuration.

## External acceptance blocker

Before merge, all eight npm packages must have trusted-publisher records requiring the exact `npm-publish` environment while retaining owner `openclaw`, repository `fs-safe`, and workflow `release.yml`. Every otherwise-equivalent environment-free connection must be removed. Missing even one platform package can cause a partial release.

The npm-side records could not be inspected from this task because the authenticated browser-control runtime failed initialization. Do not mark this PR ready or trigger a release until all eight records are verified.

## Validation

- exact Codex Astra post-fix challenge: ACCEPT for the repository half
- `git diff --check`: clean apart from checkout line-ending notices
- full remote CI: green across Linux, macOS, Windows, native, coverage, CodeQL, workflow lint, package smoke, and benchmark checks
- performance: no production/runtime hot path changes; release adds only a reviewer wait


steipete:
Closing this as an incomplete release configuration change. The existing npm-publish environment permits administrator bypass, so the proposed unconditional independent-approval guarantee is inaccurate. The corresponding trusted-publisher settings across all eight npm packages also have not been verified against the new environment binding; changing only the workflow could break publication.

This can be reopened as one coordinated change with verified environment and npm trust settings. No release, package publication, or environment-policy mutation was performed during this review.

PR #461, opened 2026-09-17
fix(file): consume async temp handles once

steipete:
## Summary

- consume each retained async atomic temp handle before awaiting its terminal `close()`
- prevent an adapter close that releases then rejects from being retried during publication recovery or a repeated terminal settlement
- preserve the existing one-attempt cleanup of a locally opened publication handle
- add owner-level, public-boundary, documentation, and real successful-path benchmark evidence

## Correctness impact

A custom adapter can release its resource and then reject its close promise. Retaining that already-consumed capability allowed recovery or a repeated `finish()` to call the same adapter handle again. Detaching before the await makes ownership exact-once while preserving failure reporting, publication bytes, identity checks, and local publication-handle cleanup.

This branch is based on `af017899d1f7045f3da2b4cc5a7583712a0f93be`. It should be rebased in landing order with #460 so falsy terminal close failures retain that PR's explicit failure-envelope semantics.

## Astra challenge

The exact patch was reviewed by Codex Astra at xhigh effort. The first challenge rejected an overbroad documentation guarantee; the text was narrowed, the bytes were re-frozen, and the exact recheck returned **ACCEPT** with no P1/P2/P3 findings.

- frozen manifest SHA-256: `ee79c31ae1683d7070703fabd317f1cb491561752ccd3425a2b9632b90764898`
- stable tracked patch ID: `711b316d2171ea1c5761b257e6b1a3cad03c0944`

## Validation and performance

- `git diff --check` passed before commit
- file line budgets pass statically
- no local Node, build, test, compiler, or benchmark command was run under the active campaign resource constraint
- hosted CI is green across the complete required matrix
- static review: no added syscall, await, integrity pass, or buffering; the successful async path adds the required retained-handle read/clear ownership transition
- a receipt-bound real public `replaceFileAtomic` benchmark includes terminal finish/close with setup and verification outside timing
- empirical strict cross-platform performance proof remains pending, so this PR stays draft


steipete:
Closing as superseded by #460 at `78453d883e3c`. The exact-once async handle-consumption change and its publication/terminal-close regressions are retained, including undefined and null rejection values. The existing 24-row settlement matrix covers the duplicate successful benchmark route.

The consolidated candidate passed the full Linux check gate and independent review. It remains subject to performance validation, CI, and the required independent code-owner approval before merge.

PR #466, opened 2026-09-17
fix(file): propagate copy fallback close failures

steipete:
Select close-error behavior from the destination writer actually opened. The fix covers an absent destination with restore-original, preserves earlier failures (including falsy values), and closes each acquired writer once. Focused source and platform regressions are retained.

Independent Astra reviewed the change and its scoped performance evidence. Shared Astra high autoreview of candidate `cf1d798f9542826dee7aa82e22921a4059eaae6e` against the preceding synchronous-admission fix at `356113cb6a10648b332c9d8937866f77f63fcbc5` was scoped-clean with no actionable P0-P2 findings. A separate static integration check confirmed both original regression suites and benchmark modules are unchanged, both validators and workload receipts are retained, and the union introduces no extra filesystem observations or descriptor calls. Full hosted Node 22/24 platform and native checks are required on the final integrated head before landing; no local build, test, or benchmark was run during this landing pass.

Final candidate `c93d3da69b0499b3b15951a23f8f40123c4a8eba` includes main `e0e4359243ed8070be8807da519e7f3dae5ff0b5`. Its complete file tree is identical to the reviewed `cf1d798` candidate. Hosted checks run against this final head.

The frozen Node 24.21.0 Linux comparison measured 12 copy-fallback settlement rows in native-off mode: async/sync, existing destinations, none/restore-original, and three payload sizes. Candidate/baseline median differences ranged from -17.1% to +7.7%; same-artifact controls ranged from -3.8% to +2.7%. The absent-destination repair is covered by correctness tests rather than these existing-destination timing rows. These are scoped original-feature measurements, not a new integrated-head measurement or acceptance of an earlier or wider campaign.


steipete:
Hosted verification update: independent Codex Astra triage found no attributable P1/P2/P3 issue and authorized one complete unchanged-head rerun after two unrelated Windows JSON-store timeout failures. Full workflow run 35225969657 attempt 2 succeeded at head 9582cfa898fd7bcf296349dbbb509b345cdd1e24; the PR now has 23 passing checks, 8 expected skips, and no failures. It remains draft pending its serialized receipt-bound performance campaign; ordinary CI is not performance acceptance.

PR #462, opened 2026-09-17
fix(permissions): preserve empty owner diagnostics

steipete:
## Summary

- preserve empty-string Windows owner-query failures as explicit fail-closed diagnostics and retain the original thrown value
- add public regressions for every falsy JavaScript rejection value and synchronous empty-string throws
- add an exact, receipt-bound Windows permission-routing workload plus a required live platform control

## Security review

Exact gpt-6-astra/xhigh post-fix review: ACCEPT, no P1/P2/P3 findings.

Reviewed staged tree: f31b4a29420e5cb0de76fb23ba083a7116ab77c1  
Stable patch ID: dc8e82e6e620e332c50ab7b18ac0aa5301128934  
Physical-byte manifest: 48ba6a57358e9f9fd675962c1df32a0af89b6b6892d40947165163b5696ef70f

## Validation

- git diff --cached --check
- hosted CI requested by this PR
- local Node/pnpm execution intentionally omitted under the active constrained-host policy

## Performance gate

No timing result is claimed yet. Keep this PR draft until the serialized cross-platform/source/control campaign completes and every predeclared cell passes the strict median and maximum-sample thresholds.

PR #467, opened 2026-09-17
fix(permissions): contain hostile diagnostics

steipete:
Keep Windows permission inspection fail-closed when an injected executor rejects with hostile objects, proxies, accessors, functions, symbols, or byte-buffer subclasses. Bound diagnostic reads without invoking user conversion hooks and preserve the original cause. Incorporate #462's explicit empty-string failure handling and regressions.

Delay wrapped-error metadata reads until a native PermissionCommandError is recognized, and reuse an immutable missing-field sentinel. Malformed wrapped errors still fall back to raw command metadata; present undefined values remain distinct from absent fields.

Validation: full Linux pnpm check, 63 focused hostile-diagnostic/byte-intrinsic/permission tests, and independent Codex review passed. The original broader Windows focused cases also passed. Fresh controlled success-path timing was 5.31 us versus 5.28 us on main (same-artifact control +0.8%); the initial mixed-cohort success slowdown did not persist after warming that row independently. Ordinary error extraction was about 6.74 versus 6.26 us, and raw Buffer stderr 8.15 versus 6.76 us. These injected cohorts expose the bounded diagnostic cost without subprocess startup. Historical broad campaign acceptance is not claimed.

Independent code-owner approval is required before merge.


steipete:
CI follow-up on exact head \85e0e0bfa9edce882ae77af6fcaa7ee286eb716\: the prior Windows Node 22/24 failures were deterministic harness admission failures—the live control forced native-required mode in ordinary fallback jobs that deliberately have no staged \.node\ helper. The correction now probes the bundled capability, skips the live control only in non-required fallback jobs, and explicitly executes it with \FS_SAFE_NATIVE_MODE=require\ after \
ative:build\ in native-check plus Windows package-smoke for Node 22/24. Runtime source and timed workloads are unchanged. Fresh exact-tree Astra post-fix review: ACCEPT. A complete new-head CI matrix is required; no selective rerun. Final performance evidence will be produced for this corrected head after the active serialized campaign completes.

PR #471, opened 2026-09-18
feat(native): keep features available without optional addons

steipete:
## What Problem This Solves

Omitting optional native packages disabled several filesystem, archive, and Windows security features.

## User Impact

Node.js consumers can use no-clobber moves, exclusive publication, retained-file staging, compressed TAR and ZIP, Windows security helpers, and copy/temp operations in native `auto` or `off` mode without an addon. Native acceleration remains preferred. Fallback warnings and receipts describe the actual mechanism; explicit global `require` remains an addon-availability diagnostic.

Permission, identity, unsafe-input, cancellation, and I/O failures still reject. Portable pathname checks retain their documented concurrency limits. A failed move can leave a published target or retained source; completion and recovery details describe that outcome without granting cleanup authority. Filesystems without hardlinks use atomic platform commands: Linux requires isolated system Python 3 with `ctypes` and libc `renameat2`, macOS uses JXA, and Windows uses PowerShell/.NET. These command routes warn about startup overhead.

## Why This Change Was Made

Select guarded alternatives by the individual missing capability. Portable moves keep the original inode and publish destinations exclusively. POSIX captures and verifies the source before removal; Windows retires it through a verified source-name handle. Observed substitutes are preserved, confirmed completion survives later verification/close errors, and uncertain operations are never retried or rolled back automatically.

Archives use required lazy JSZip and bundled, bounded bzip2/zstd WASM decoders. The change preserves physical ZIP metadata, rejects corrupt declared-zero entries and malformed zstd Huffman streams, and fixes a Node 22 decoder teardown hang while preserving the original error. Requested cloning and bounded cleanup use portable equivalents with truthful warnings and receipts.

## Evidence

- Node 22.23.2 and Node 24.21.0 each passed 12,467 tests, with 571 platform/capability skips. Node 24 `pnpm check` passed, including build, documentation examples, and package checks. The unchanged Rust production sources passed 123 tests.
- Fresh npm 11.19.0 and pnpm 11.25.0 installations passed with and without optional packages. Four optional-free consumers in `off`/`auto` imported all 23 code subpaths without loading an addon; their 38 checked compiled-module hashes matched the build, including archive and source-replacement scenarios.
- The public-surface source audit covers 24 Root methods and 226 distinct runtime exports. No ordinary missing-addon gate remains in `auto`/`off`; this is not an exhaustive per-export execution or benchmark claim.
- Managed review covered the full change, followed by a clean review of the corrections. The command-completion regression failed before its fix and passed on both Node versions.
- The final tarball passed ordinary Linux and real FAT32 cases in both `off` and `auto`, including twelve successful calls through the shipped isolated Python rename bridge. The cases verify collisions, complete staged/streamed publication, inode/mode/content preservation, independent copies, and owned fixture cleanup.
- Committed Windows validation and PR CI are pending; this evidence section will be updated with their actual results before merge.


steipete:
Closing this version after the maintainer review. Optional-free operation is useful, but this proposal changes existing explicit guarantees: `cleanupSafety: "require-bounded"` can select recursive pathname cleanup, `clone: "always"` becomes best-effort copying, and a loaded addon in native `require` mode can fall back when an individual retained-directory primitive is unavailable. Warnings or new mechanism fields do not preserve those caller contracts.

The 220-file change also combines archive codecs, Windows security commands, and move/staging/cleanup policy, which need separate correctness and performance decisions. Native availability should improve without redefining strict options. A focused follow-up can retain the useful portable mechanisms under explicit compatible policies, keep strict modes fail-closed, and measure command startup and fallback copying costs. The archive work can be reviewed independently.

This is a decline of the current design, not a claim that the implementation is obsolete. The branch and review history remain available for extracting the compatible pieces. Current-head CI also has unresolved failures, so the proposal is not ready to land as submitted.

PR #473, opened 2026-09-18
feat(copy): read APFS clone metadata without native addons

steipete:
`readCloneFileMetadata()` currently requires the native addon even on macOS, where the same filesystem query is available through the built-in system runtime. This adds a bounded JXA fallback for `auto`/`off`, retaining exact 64-bit clone/inode identifiers, input order, existing path normalization, and per-file unsupported results. Known non-macOS platforms return the same unsupported observations as the native implementation.

Paths travel as JSON data to a fixed program, in batches capped at 128 paths and 64 KiB. Output and execution time are bounded, failed input transport joins child cleanup, and command failures reject. The fallback matches Node/N-API's replacement of malformed UTF-16 in filenames. Explicit native `require` still rejects a missing addon or reader; native operational errors are never retried. Clone creation/copy policies and bounded cleanup are unchanged.

Native remains the fast path. Synthetic macOS arm64 measurements put command startup around 52–54 ms for one file, while a batch of 128 takes about 68–69 ms. The native measurements were below 1 ms for those inputs. A deduplicated, path-free warning explains this cost. The method benchmark now exercises this route instead of skipping it without an addon.

Validation:

- Full `pnpm check`: 9,572 passed, 2,566 platform/feature skips; build, docs, lint, API and package checks passed.
- Focused Node 22.23.2 and 24.21.0 suites: 106 passed and 8 existing skips each, including all 92 added cases.
- Real native/portable comparisons on APFS cover clone identities, large ordered batches, independent copies, symlinks, missing paths, permissions, unusual filenames and malformed UTF-16.
- Actual npm install of the clean committed package with `--omit=optional`: `off` and `auto` returned exact file/clone identities, independent copies remained distinct, malformed-surrogate and missing-entry behavior matched, `require` rejected, and no `.node` addon loaded. The receipt binds package integrity and the three changed compiled modules.
- Managed independent reviews found no actionable P0–P2 findings. The portable method benchmark also passed.

Runtime equivalence was exercised on macOS arm64/APFS; Intel macOS and non-APFS live execution were not tested locally. All 24 required exact-head checks pass in [CI](https://github.com/openclaw/fs-safe/actions/runs/35302674882), including native/platform/package coverage. GitHub has automatically requested the required code-owner review before merge. No release or version change is included.


steipete:
Closing this proposal for the current consolidation pass. The existing native metadata API remains supported. This adds a separate JXA/Objective-C subprocess implementation for installations that omit the available macOS addon, with roughly 50–70 ms process-startup cost and no identified production caller in the proposal.

The bounded review did not find a concrete strict-clone or filesystem-boundary defect. The decision is about keeping the implementation and supported mechanisms small: this extra backend is not justified for the current maintenance scope. A future proposal should establish a concrete consumer that cannot use the native package and address the comparison-harness admission and APFS-result verification gaps described by the code review. The branch and proof history remain available.

PR #478, opened 2026-09-18
feat(root): support no-clobber moves without native addon

steipete:
OpenClaw's npm recovery path installs with [`--omit=optional`](https://github.com/openclaw/openclaw/blob/3c4c111b0f71d00f7f94a9f31d5030436370aa4f/src/infra/update-global.ts#L1409), but its [outgoing-media spool](https://github.com/openclaw/openclaw/blob/3c4c111b0f71d00f7f94a9f31d5030436370aa4f/src/infra/outbound/delivery-queue-media-spool.ts#L132), [migration source claims](https://github.com/openclaw/openclaw/blob/3c4c111b0f71d00f7f94a9f31d5030436370aa4f/src/infra/state-migrations.source-snapshot.ts#L107), and [plugin-tree repair snapshots](https://github.com/openclaw/openclaw/blob/3c4c111b0f71d00f7f94a9f31d5030436370aa4f/src/infra/update-candidate-plugin-tree.ts#L574) use no-clobber `Root.move()`. Without the optional addon those required moves currently fail with `helper-unavailable`.

Allow `Root.move()` in `auto` and `off` to use a true platform no-replace rename command when a native capability is missing before admission. Linux uses isolated Python standard-library ctypes with retained descriptors; macOS uses the built-in JXA host with retained parent descriptors; Windows uses fixed readable package assets with retained metadata handles. Normal Windows script policy and endpoint protection remain in force. Native dispatch remains preferred and unchanged, strict `require` rejects missing capabilities, and native errors never trigger a command retry. Standalone publication and retained-directory staging keep their current native requirements.

Snapshot merged Root.move mutation policy before asynchronous admission, and retain live authority callbacks at dispatch. The portable owner retains Root confinement, source identity, parent custody, and post-operation checks. Portable receipts distinguish explicit pre-dispatch rejection from an indeterminate outcome and a confirmed rename. Every ordinary errno/NTSTATUS after dispatch, including collisions, remains indeterminate; no hardlink/copy/source-removal emulation or cleanup of uncertain state is introduced. A one-time warning explains command startup and best-effort name-swap limits.

Validation against base `59f297b1` and the frozen candidate:

- Fresh 0.15.0 Darwin native build, 285 focused tests, package/API smoke, docs examples, file-size/boundary lint, and whitespace checks passed.
- Linux x64/glibc: Node 22.23.2 and 24.21.0 each passed 313 focused tests (5 expected platform skips). Fresh `--omit=optional --ignore-scripts` installed consumers passed 284 public filesystem cases in total across both runtimes on real ext4 and FAT32, including separately labelled native controls. Main reproduces the missing-helper failure before the fix.
- Windows: the unchanged bridge passed 9 case-sensitivity cases on each Node version, covering exact root/parent/leaf identities and case-distinct targets in `off`, missing-addon `auto`, and native `require`, plus ordinary case-insensitive alternate-spelling controls. This agrees with [Microsoft's per-directory case-sensitivity contract](https://devblogs.microsoft.com/commandline/per-directory-case-sensitivity-and-wsl/).
- Before main integration, the complete check passed 13,708 tests (583 skipped) on `46efdfda`; current-base validation is the focused/package proof above.

The broader Windows Root.move suite is still pending. Linux proof covers x64/glibc with installed Python3; other Linux architectures have UAPI/selection tests but no hardware proof here. Native comparison results are recorded individually and do not claim identical behavior for arbitrary callback mutations.


steipete:
Closing in favor of the matching-prebuild installer repair merged in https://github.com/openclaw/openclaw/pull/152349.

The concrete OpenClaw failure was its npm recovery path omitting optional dependencies, which could leave the matching fs-safe addon absent. The installer repair checks the addon and restores the exact platform/version package when it is missing. Existing proof covers the real published updater with an explicit missing-prebuild fault, Node 22/24 lifecycle paths, macOS package-manager shims, and Windows lifecycle/deferred-install paths. The restored addon successfully performed native operations, including no-clobber moves and collision preservation in the Linux updater proof.

I also rechecked the actual npm tarballs for all seven native targets at both 0.13.1 (used by published OpenClaw) and 0.15.0 (current fs-safe): all 14 packages are available, have verified registry integrity, correct OS/CPU/libc metadata and binary architecture, and contain the N-API registration symbol. This artifact check is separate from runtime proof and is not a guarantee for unsupported architectures, explicitly disabled native mode, corrupt installations, or loader restrictions.

Release status matters: OpenClaw 2026.9.5 on npm still contains the old installer and does not contain the repair helper. The fix is merged on main and needs to ship in the next OpenClaw release. This closure does not claim that existing released installations have already received that repair.

This PR's Python/JXA/PowerShell route would establish a separate capability for intentionally addon-free no-clobber moves; we are choosing the native-prebuild installation strategy for the reported problem. The independent Root.move mutation-policy fix already landed in #479. The additional Linux O_PATH ABI coverage can be extracted separately if support beyond the published native targets is wanted.

PR #507, opened 2026-09-19
perf(archive): accelerate bounded ZIP end-record scans

steipete:
## What Problem This Solves

ZIP admission scans every byte of the bounded 65 KiB end-record window in JavaScript, even for ordinary archives with a single end record.

## User Impact

Reduce CPU work when loading, extracting, or reading ZIP archives while preserving all framing, comment-length, ambiguity, and byte-bound checks.

Follow-up commits relocate the unchanged release note, synchronize current main for required checks, and give two slow Windows I/O test fixtures bounded 30-second deadlines. The ZIP optimization and its benchmark harness are unchanged; fixture payload sizes, all assertions, and non-Windows deadlines are preserved.

## Why This Change Was Made

Use native buffer signature search for sparse end records. After 16 candidate signatures, use the original bounded byte scan so dense attacker-controlled payloads and comments retain predictable cost. Public APIs, errors, and defaults are unchanged.

## Evidence

Measured baseline: `6953a1d85a92160f036b9ce9cfd5c8f096f92cbe`; candidate runtime: `dbf25fb606235ca3e12526248b0986f669685322`.

- Focused ZIP admission and metadata suites: 108 passed, 60 platform/native skips.
- Added coverage for empty and maximum comments, truncation, short input, dense false signatures, and ambiguous end records.
- Isolated Codex review through P2: no actionable findings; `git diff --check` passes.
Paired Node 24.21 macOS arm64 public-loader benchmarks used native mode `off`, ABBA then BAAB order, nine samples per run, and 36 pooled samples per version. Ordinary cohorts use 300 calls/sample; 512-entry cohorts use the same harness divisor for both versions. Reports share one harness hash and exactly two distribution hashes.

| Complete public loader workload | Baseline → candidate (µs/call) | Speedup |
| --- | ---: | ---: |
| 64 KiB stored payload | 164.46 → 65.12 | 2.53× |
| Maximum-length comment | 276.95 → 175.42 | 1.58× |
| Dense false signatures | 182.42 → 184.53 | 1.16% slower |

All 13 rows completed successfully. The tiny ZIP control moved by +0.63 µs with overlapping spread; 512-entry rows improved by 1.7–5.8%, too small for a broad scaling claim. The capped search avoids the substantial dense-marker regression seen with uncapped native search.

Complete local `pnpm check` passed again after synchronization with current main on Node 24.21: 543 test files passed (10,957 tests), with 41 files/3,524 tests skipped for unavailable platforms or native bindings. Lint, build, WASM allocator proof, documentation examples, and installed-package checks also passed.

The first Windows Node 24 CI attempt timed out in an unchanged 17 MiB descriptor-copy test; that exact job passed when retried with identical source and settings.

A later Windows Node 22 run timed out at 5.229 seconds in the unchanged multi-operation Root-defaults scenario. That case and the earlier large-copy case now use bounded 30-second Windows-only test deadlines, matching existing large-I/O proof fixtures. Both focused suites passed and the full local gate passed again.


steipete:
Follow-up on the late Node 26 question raised by the historical #388 results: the reported 8–20% many-entry regression did not reproduce with the current parent/candidate on a separate ARM64 Mac.

Compared baseline `fe15b87925ade96af6bba249409d86e2ebef5cb1` with merged candidate `13dd9e7a4d58594cba4c200ca9f524ee39b2ae84` using Node 26.8.2 on macOS ARM64, Apple M4 Pro, native mode `off`, and identical JSZip 3.10.2 dependencies. Every built JavaScript file matched except `archive-zip-directory.js`. Artifact and dependency hashes were verified before and after execution.

A narrow public-loader harness reproduced the repository's exact 512-entry STORE/DEFLATE fixtures and 2,048-entry ASCII fixtures at depths 1 and 8. Fresh processes ran source ABBA, an identical-artifact ABBA control, then source BAAB. Each process used 20 warmups and seven samples, with 100 calls/sample for 512 entries and 10 for 2,048. Payload and entry-count assertions ran outside timing.

| Public loader workload | Baseline median, µs | Candidate median, µs | Pooled change |
| --- | ---: | ---: | ---: |
| 512 ASCII, STORE | 2,190.30 | 2,133.31 | -2.60% |
| 512 ASCII, DEFLATE | 2,193.15 | 2,132.21 | -2.78% |
| 2,048 ASCII, depth 1 | 9,855.43 | 9,584.52 | -2.75% |
| 2,048 ASCII, depth 8 | 24,479.96 | 23,901.64 | -2.36% |

Individual source-block changes ranged from -7.39% to +2.26%; identical-artifact controls ranged from -2.85% to +2.13%. These results support parity for these Node 26 many-entry cases, with no substantial speedup claim. A separate interleaved paired experiment was consistent with that conclusion.

Initial shared-workstation timings were unsuitable for inference: identical-artifact controls varied by more than 50% under heavy host contention. Those measurements were retained but were not used to claim either a regression or a gain. No production code was changed in response to the noisy run. The substantial speedup claims in this PR remain scoped to the measured Node 24 large-payload and maximum-comment workloads.

